PR #34, opened 2023-05-03
Implement functionality for importing wallets from the genesis and browser wallet key export formats.

vikt0r0:
## Purpose

Implement functionality for importing wallets from the genesis and browser wallet key export formats.

## Changes

- Add `WalletAccount` similar to how it is in the rust SDK.
- Add functionality for creating a `WalletAccount` instances from the genesis and browser wallet key export formats. 
- Implement tests.

## Checklist

      hard-to-understand areas.
- [X] (If necessary) I have updated the CHANGELOG.


src/ConcordiumNetSdk/Types/AccountAddress.cs
```diff
@@ -6,12 +6,15 @@ namespace ConcordiumNetSdk.Types;
...
 ///
-/// The address of an account is a 256-bit value which uniquely identifies the account.
+/// The address of an account is a 32-byte value which uniquely identifies the account.
 /// </summary>
 public readonly struct AccountAddress : IEquatable<AccountAddress>
 {
     public const UInt32 BytesLength = 32;
 
+    /// <summary>
+    /// A version byte that is prepended to addresses represended as base58 strings.
```
abizjak, line 16:
```suggestion
    /// A version byte that is prepended to addresses represented as base58 strings.
```

src/ConcordiumNetSdk/Wallets/Json/AccountKeys.cs
```diff
@@ -0,0 +1,74 @@
...
+///
+/// Such can be parsed into an instance of this class using
+/// <see cref="Newtonsoft.Json.JsonConvert"/>.
+///
+/// The object indexed by this field holds the sign keys and can
+/// be parsed into a map from pairs of <see cref="AccountCredentialIndex"/>
+/// and <see cref="AccountKeyIndex"/>es to <see cref="Ed25519SignKey"/>s
+/// using <see cref="TryGetSignKeys"/>.
+/// </summary>
+public class AccountKeys
```
abizjak, line 19: Does this need to be public?

src/ConcordiumNetSdk/Wallets/Json/BrowserWalletExportFormat.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    public Value? value { get; set; }
+
+    public AccountAddress TryGetAccountAddress()
+    {
+        try
+        {
+            if (value is null)
+            {
+                throw new ArgumentNullException("Required field 'value' is null.");
```
abizjak, line 23: Wouldn't it be better to say "not present" instead of "is null"?

src/ConcordiumNetSdk/Wallets/Json/KeyInfo.cs
```diff
@@ -0,0 +1,15 @@
...
+
+/// <summary>
+/// Represents (a subset of) a JSON object indexed by the
+/// <c>keyInfo</c> field of the browser and genesis wallet
+/// export JSON formats.
+///
+/// Such can be parsed into an instance of this class using
+/// <see cref="Newtonsoft.Json.JsonConvert"/>.
+/// </summary>
+public class KeyInfo
```
abizjak, line 11: Does this need to be public?

src/ConcordiumNetSdk/Wallets/Json/Key.cs
```diff
@@ -0,0 +1,15 @@
...
+
+/// <summary>
+/// Represents (a subset of) a JSON object indexed by the
+/// <c>key</c> field of the browser and genesis wallet
+/// export JSON formats.
+///
+/// Such can be parsed into an instance of this class using
+/// <see cref="Newtonsoft.Json.JsonConvert"/>.
+/// </summary>
+public record Key
```
abizjak, line 11: Does this need to be public?

src/ConcordiumNetSdk/Wallets/Json/Value.cs
```diff
@@ -0,0 +1,18 @@
...
+
+/// <summary>
+/// Represents (a subset of) a JSON object indexed by the
+/// <c>value</c> field of the browser and genesis wallet
+/// export JSON formats.
+///
+/// Such can be parsed into an instance of this class using
+/// <see cref="Newtonsoft.Json.JsonConvert"/>.
+/// </summary>
+public class Value
```
abizjak, line 14: Also this.

Do these need to be separate classes? In that case why are they not records?

tests/ConcordiumNetSdk.UnitTests/EmbeddedResourceDataAttribute.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using Xunit.Sdk;
+
+/// <summary>
+/// Data attribute which allows to capture data of resources embedded
+/// in the manifest for use in tests.
```
abizjak, line 9: I don't understand this comment. What is the "data attribute"?

tests/ConcordiumNetSdk.UnitTests/Wallets/Data/BrowserWalletKeyExportFormatValid.json
```diff
@@ -0,0 +1,25 @@
+{
+    "type": "concordium-browser-wallet-account",
+    "v": 0,
+    "environment": "testnet",
+    "value": {
+      "accountKeys": {
+        "keys": {
+          "0": {
+            "keys": {
+              "0": {
```
abizjak, line 10: Can we try with index 17 and 37 as well please.

src/ConcordiumNetSdk/Wallets/WalletAccount.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// <summary>
+    /// Try to create a new instance from a JSON string in the browser wallet key export format.
+    /// </summary>
+    /// <param name="json">JSON string in the browser wallet key export format.</param>
+    /// <exception cref="JsonException">The specified input is not valid JSON.</exception>
+    /// <exception cref="WalletDataSourceException">Either a field is missing or an index or sign key could not be parsed.</exception>
+    public static WalletAccount FromBrowserWalletExportFormat(string json)
+    {
+        Json.BrowserWalletExportFormat genesisWallet =
+            JsonConvert.DeserializeObject<Json.BrowserWalletExportFormat>(json);
```
abizjak, line 93: So in the Rust SDK we just have a function where you give it a file or JSON Value and it will figure out what to do.

Here the user has to explicitly choose what they have which makes it more complex to use this, i.e., to make a service which takes any keys.

Is it not possible to do the same as we do in the Rust SDK?

I feel that the interface IWalletDataSource is not really a good solution to this. That should at best be an implementation detail.

Is it not possible to hide this like we do in the Rust SDK? Are there limitations that prevent that?

src/ConcordiumNetSdk/Wallets/Json/Value.cs
vikt0r0: Opposed to several classes do you mean that they could be nested?

src/ConcordiumNetSdk/Wallets/Json/Key.cs
vikt0r0: Do you believe it should be `internal`?

src/ConcordiumNetSdk/Wallets/Json/KeyInfo.cs
vikt0r0: Do you believe it should be `internal`?

src/ConcordiumNetSdk/Wallets/Json/AccountKeys.cs
vikt0r0: Do you believe it should be `internal`?

tests/ConcordiumNetSdk.UnitTests/EmbeddedResourceDataAttribute.cs
vikt0r0: That's the Xunit lingo for a data source for a data theory. It refers to the super class ("is-a"). I will link it in this doc, however since it may be confusing.

src/ConcordiumNetSdk/Wallets/Json/Value.cs
abizjak: Yes. I know this is a Java/C# thing but I find these scattered classes in all the different files rather hard to keep track of.

src/ConcordiumNetSdk/Wallets/Json/Key.cs
abizjak: As a general rule the smaller the API surface is the better since it is easier to find what you need.

And if these keys are only internal then I would make them as such.

tests/ConcordiumNetSdk.UnitTests/EmbeddedResourceDataAttribute.cs
abizjak: Ok, thanks.

src/ConcordiumNetSdk/Wallets/Json/AccountKeys.cs
vikt0r0: I made it internal :-)

src/ConcordiumNetSdk/Wallets/Json/Value.cs
vikt0r0: Yea, I agree. I nested as much as possible.

src/ConcordiumNetSdk/Wallets/WalletAccount.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            try
+            {
+                return FromBrowserWalletKeyExportFormat(json);
+            }
+            catch (Exception)
+            {
+                throw e;
+            }
+        }
```
abizjak, line 93: I'm usually not a fan of this approach. In particular the browser wallet export explicitly has
```
  "type": "concordium-browser-wallet-account",
```
there to distinguish it from others. So ideally we'd dispatch on that rather than "try and catch" way of doing it.

abizjak (approved):
I would prefer that the parsing from JSON was clever in detecting the format, rather than try and catch.

But if that's too complex to implement we can just go with the current solution.

PR #35, opened 2023-05-04
Add examples and helpers motivated by these to SDK.

vikt0r0:
## Purpose

Adds some examples that show how to use the SDK. This should comprise runnable examples with behaviour similar to [examples in the Rust SDK](https://github.com/Concordium/concordium-rust-sdk/tree/main/examples). Specifically examples are added for the supported transactions in the current iteration of the SDK, namely `Transfer`, `TransferWithMemo` and `RegisterData`. Furthermore examples on how to work with "raw" methods generated from the GRPC API V2 definition are added, specifically `GetAccountInfo`, `GetBlockTransactionEvents` and the streaming call `GetFinalizedBlocks`.

The client was refactored a bit to improve on overall structure and helper methods `SendTransaction` and `GetNextAccountSequenceNumber` since examples could benefit from the ergonomics they provide.

## Changes

- Added "scaffolding" for examples by providing helper classes that allow for specifying command line parameters for examples.
- Implemented some helper functions to `ConcordiumClient` for sending transactions and retrieving account sequence numbers when working with transactions.
- Factor out wrappers for raw methods into `RawClient` class. Add `Raw` field of this type to `ConcordiumClient` s.t. the "raw" GRPC types methods are "namespaced" and can be accessed by `ConcordiumClient.Raw.MethodName(...)`.
- Implemented examples for working with `Transfer`, `TransferWithMemo` and `RegisterData` transactions as well as invoking raw client calls `GetAccountInfo`, `GetBlockTransactionEvents` and `GetFinalizedBlocks`.

## Checklist

- [ ] I have performed a self-review of the changes.
      hard-to-understand areas.


src/ConcordiumNetSdk/Client/RawClient.cs
```diff
@@ -0,0 +1,841 @@
...
+using Grpc.Core;
+using Grpc.Net.Client;
+
+namespace ConcordiumNetSdk.Client;
+
+/// <summary>
+/// The "raw" GRPC client interface exposing wrappers that abstract away
+/// connection handling and parameters present in generated methods.
+/// </summary>
+public class RawClient : IDisposable
```
vikt0r0, line 11: I guess this could technically be suited as a nested class of `ConcordiumClient`, however it was a bit hard on the eyes IMO, so I factored it out.

abizjak: Ok, but could we make it not public??

vikt0r0: Definitely.

src/ConcordiumNetSdk/Types/BlockHash.cs
```diff
@@ -52,4 +57,34 @@ public Concordium.V2.BlockHash ToProto()
...
+    {
+        return new BlockHashInput() { Best = new Empty() };
+    }
+
+    /// <summary>
+    /// Returns a <see cref="BlockHashInput"/> which specifies the last finalized block.
+    ///
+    /// This can be used as the input for class methods of <see cref="ConcordiumClient.RawClient"/>.
+    /// </summary>
+    public static BlockHashInput LastFinalBlockHashInput()
```
abizjak, line 86: Shouldn't we call this just LastFinal? Same for Best? Don't we call this as


```
BlockHashInput.LastFinal
```

examples/ConcordiumNetSdk.Examples.Transactions.RegisterData/Program.cs
```diff
@@ -0,0 +1,64 @@
...
+///
+/// Cfr. <see cref="RegisterDataTransactionExampleOptions"/> for more
+/// info on how to run the program, or refer to the help message.
+/// </summary>
+class Program
+{
+    static void SendRegisterDataExample(RegisterDataTransactionExampleOptions options)
+    {
+        // Read the account keys from a file.
+        WalletAccount account = WalletAccount.FromBrowserWalletExportFormat(options.WalletKeysFile);
```
vikt0r0, line 25: These are not tested, and I just noticed that these examples currently do not read the file, but pass the specified path instead. I'll fix it and test them before requesting re-review and merging.

examples/ConcordiumNetSdk.Examples/ExampleOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+/// This class can be used to specify options for the
+/// <see cref="ConcordiumNetSdk.Client.ConcordiumClient"/>
+/// at the command-line.
+/// </summary>
+public abstract class ExampleOptions
+{
+    /// <summary>
+    /// Default URL representing the endpoint where the GRPC V2 API is served.
+    /// </summary>
+    const string DEFAULT_URL = "//localhost";
```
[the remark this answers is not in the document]
vikt0r0: It's due to the current weird handling of schemes, but I agree we should change things according to your other comment. I'll fix it now.

src/ConcordiumNetSdk/Types/BlockHash.cs
vikt0r0: I can add classes `BlockHashInput` and `AccountIdentifierInput` if you want me to. Currently these are instantiated from `BlockHash` and `AccountAddress` 🤔 but it will probably me more correct to have these and then invoke `ToProto` with their instances as the receiver?

vikt0r0: It's perhaps better to scrap these for now and do `BlockHashInput() { Best = new Empty() }`. It doesn't really seem right to me to do it this way.

examples/ConcordiumNetSdk.Examples.Transactions.TransferWithMemo/Program.cs
```diff
@@ -0,0 +1,65 @@
...
+        ConcordiumClient client = new ConcordiumClient(
+            new Uri(options.Endpoint), // Endpoint URL.
+            options.Port, // Port.
+            60 // Use a timeout of 60 seconds.
+        );
+
+        // Create the transfer transaction.
+        CcdAmount amount = CcdAmount.FromCcd(options.Amount);
+        AccountAddress receiver = AccountAddress.From(options.Receiver);
+        // Encode a singleton string as CBOR and use that for the memo data.
```
abizjak, line 35: What is a "singleton" string?

examples/ConcordiumNetSdk.Examples/Example.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// </param>
+    /// <param name="exampleCallback">
+    /// The callback corresponding to the example program which will be
+    /// invoked with the parsed <typeparam name="T"/> instance as its
+    /// argument.
+    /// </param>
+    public static void Run<T>(string[] args, Action<T> exampleCallback)
+        where T : ExampleOptions
+    {
+        /// This is a convenience methos which leverages to other overload
```
abizjak, line 32:
```suggestion
        /// This is a convenience method which leverages to other overload
```

src/ConcordiumNetSdk/Client/ConcordiumClient.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// <returns>The best guess of the next sequence number for the supplied account.</returns>
+    /// <exception cref="RpcException">The call failed.</exception>
+    public ConcordiumNetSdk.Types.AccountSequenceNumber GetNextAccountSequenceNumber(
+        ConcordiumNetSdk.Types.AccountAddress accountAddress
+    )
+    {
+        NextAccountSequenceNumber next = Raw.GetNextAccountSequenceNumber(accountAddress.ToProto());
+
+        // Return the sequence number as a "native" SDK type.
+        return ConcordiumNetSdk.Types.AccountSequenceNumber.From(next.SequenceNumber.Value);
```
abizjak, line 114: I think this should not throw away the information about whether all transactions are finalized or not.

abizjak (approved):
Looks good, but please don't throw away information in the next account sequence number query.

PR #38, opened 2023-05-12
Update endpoint GetBlockItemStatus on .NET SDK

schwartz-concordium:
## Purpose

Extend `concordium-net-sdk` such that transaction summary is easily presented to the user.

## Changes

Extended client with `GetBlockItemStatus` from a given transactionshash.

Added `BlockItemSummary` and `TransactionStatus` helper classes which makes it easier for the user to work with responses from the client.

Also refactored examples to have a simpler representation in solution and project.

## Checklist

      hard-to-understand areas.
- [X] (If necessary) I have updated the CHANGELOG.

## CLA acceptance

- [X ] I accept the above linked CLA.

## References
 - BlockItemStatus to TransactionStatus
[concordium-rust-sdk/conversions.rs at 671c84a27cb3d3b31f90fcae74e0cb88a25fe864 · Concordium/concordium-rust-sdk](https://github.com/Concordium/concordium-rust-sdk/blob/671c84a27cb3d3b31f90fcae74e0cb88a25fe864/src/v2/conversions.rs#L911) 

- TransactionStatus
[TransactionStatus in concordium_rust_sdk::types - Rust](https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/types/enum.TransactionStatus.html) 

 - BlockItemSummary
[BlockItemSummary in concordium_rust_sdk::types - Rust](https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/types/struct.BlockItemSummary.html) 
[concordium-rust-sdk/mod.rs at 671c84a27cb3d3b31f90fcae74e0cb88a25fe864 · Concordium/concordium-rust-sdk](https://github.com/Concordium/concordium-rust-sdk/blob/671c84a27cb3d3b31f90fcae74e0cb88a25fe864/src/types/mod.rs#L1033)

- [Confluence (require access)](https://concordium.atlassian.net/wiki/spaces/~7120205f63b1fc37e84c4f81189e85ff7dfc4a/pages/1134002177/Update+endpoint+GetBlockItemStatus+on+.NET+SDK)


examples/Common/Common.csproj
```diff
@@ -5,11 +5,9 @@
     <TargetFramework>net6.0</TargetFramework>
```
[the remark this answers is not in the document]
3schwartz: net7 isn’t long term support
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

I do think that users normally use LTS version. Since we are a client SDK our users would have to use the same versions as us, and it would give conflicts if we are 7 and they normally are 6.

What are you experience when using SDK’s?

src/Types/TransactionStatus.cs
```diff
@@ -0,0 +1,80 @@
+using Concordium.Grpc.V2;
+using Concordium.Sdk.Exceptions;
+
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// A common return type for TransactionStatus.
+/// </summary>
+public interface ITransactionStatus
```
vikt0r0, line 9: Are there any drawbacks from doing `TransactionStatus<T> where T : TransactionStatus<T>` and its variants `TransactionStatusCommitted : TransactionStatus<TransactionStatusCommitted>`?

Does using the interface not entail that one has to match on the type at runtime?

src/Types/RejectReason.cs
```diff
@@ -0,0 +1,421 @@
...
+  /// Account is not a delegation account.
+  /// </summary>
+  public AccountAddress? NotADelegator { get; init; }
+  /// <summary>
+  /// Delegation target is not a baker.
+  /// </summary>
+  public BakerId? DelegationTargetNotABaker { get; init; }
+
+
+  internal RejectReason(Concordium.Grpc.V2.RejectReason other)
```
vikt0r0, line 101: Document, and I'd prefer using `From` here as well for consistency.

src/Types/TransactionStatus.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            this.States.Add(state);
+        }
+    }
+}
+
+/// <summary>
+/// Creates a TransactionStatus dependent of the status in the returned BlockItemStatus from the node.
+/// </summary>
+internal static class TransactionStatusFactory
```
vikt0r0, line 69: Is it not possible to have a `From` method with the approach detailed above?

src/Types/RejectReason.cs
```diff
@@ -0,0 +1,421 @@
+using Concordium.Sdk.Exceptions;
+using Google.Protobuf;
+
+namespace Concordium.Sdk.Types;
```
vikt0r0, line 4: Is it not possible to have `RejectReason<T> where T : RejectReason<T>`? This will carry type information of the reject reason, no?

vikt0r0 (changes requested):
Looks good overall! Nice job 👍🏼 some points I think should be addressed:

- I think the example program should be consistent with the existing ones for the reasons provided in the comments.
- There should be tests added for the types where it makes sense.
- Some comments lack return values, exceptions and parameter documentation.
- A lot of files were moved for some reason. This should be reverted, or done in a separate PR if there is a good reason for doing so.
- I am a fan of leveraging type information where possible. This is already done for existing records/structs/classes, and we should probably do this unless there is a drawback with that approach?
- The formatting should be fixed in accordance with the .editorconfig file. You can invoke `dotnet format` from the command line to run it manually :-)

ConcordiumNetSdk.sln
```diff
@@ -7,19 +7,23 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Concordium.Sdk", "src\Conco
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Concordium.Sdk.Tests.UnitTests", "tests\Tests\UnitTests\Concordium.Sdk.Tests.UnitTests.csproj", "{FA99FA9D-C722-4BD4-BB6E-C83843EC5BA0}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Concordium.Sdk.Examples.RawClient.GetAccountInfo", "examples\Examples\RawClient\GetAccountInfo\Concordium.Sdk.Examples.RawClient.GetAccountInfo.csproj", "{B5E9C218-21CD-42E9-BB37-1D7AB3BAB2C6}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "examples", "examples", "{D86A7796-6AD5-48CA-9DC6-415CC53629B6}"
```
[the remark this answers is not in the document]
schwartz-concordium: I refactored the example structure due to
- It has folder debt not needed
- When opening your solution it looks more clean to have it in a nested structure, like it is in the file system. Without it’s difficult to see what is SDK and what is example - see example

Screenshot 2023-05-16 at 11 47 40

- I shortened the name since I felt it was redundant to have “Example” in names and also the SDK name

ConcordiumNetSdk.sln
```diff
@@ -7,19 +7,23 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Concordium.Sdk", "src\Conco
...
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Concordium.Sdk.Examples.RawClient.GetBlockTransactionEvents", "examples\Examples\RawClient\GetBlockTransactionEvents\Concordium.Sdk.Examples.RawClient.GetBlockTransactionEvents.csproj", "{0E117FB5-D225-40F1-AB56-BF92350E573B}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GetBlockItemStatus", "examples\GetBlockItemStatus\GetBlockItemStatus.csproj", "{CAC92E03-8FC8-474A-93F0-B31119CEE0BF}"
```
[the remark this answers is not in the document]
schwartz-concordium: See above

ConcordiumNetSdk.sln
```diff
@@ -7,19 +7,23 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Concordium.Sdk", "src\Conco
...
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Concordium.Sdk.Examples.RawClient.GetFinalizedBlocks", "examples\Examples\RawClient\GetFinalizedBlocks\Concordium.Sdk.Examples.RawClient.GetFinalizedBlocks.csproj", "{F3AED5FA-4BB6-4ADC-A155-C94A2F7A2C1F}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Common", "examples\Common\Common.csproj", "{DEFCB599-F5A6-4CD1-806A-4F77BC378562}"
```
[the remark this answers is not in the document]
schwartz-concordium: See above

ConcordiumNetSdk.sln
```diff
@@ -39,39 +43,53 @@ Global
...
-		{506A963D-B243-4DE8-B04C-7A30A2A7AAA7}.Release|Any CPU.Build.0 = Release|Any CPU
-		{E3006800-82A4-495E-8886-BC98F858875C}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{E3006800-82A4-495E-8886-BC98F858875C}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{E3006800-82A4-495E-8886-BC98F858875C}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{E3006800-82A4-495E-8886-BC98F858875C}.Release|Any CPU.Build.0 = Release|Any CPU
-		{F5B375A5-F7A7-4D01-A24F-62F4DA9D7659}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{F5B375A5-F7A7-4D01-A24F-62F4DA9D7659}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{F5B375A5-F7A7-4D01-A24F-62F4DA9D7659}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{F5B375A5-F7A7-4D01-A24F-62F4DA9D7659}.Release|Any CPU.Build.0 = Release|Any CPU
+		{CAC92E03-8FC8-474A-93F0-B31119CEE0BF}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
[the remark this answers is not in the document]
schwartz-concordium: See above

.gitignore
```diff
@@ -390,6 +390,7 @@ FodyWeavers.xsd
 **/.idea/**/*.iml
 **/.idea/**/contentModel.xml
 **/.idea/**/modules.xml
+**/.idea
```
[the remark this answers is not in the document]
schwartz-concordium: already removed

examples/Common/Common.csproj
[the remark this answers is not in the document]
ali-concordium: That's fair enough. Let's wait till November and update to version 8 👍

src/Types/ContractInitializedEvent.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <summary>
+    /// A contract name.
+    /// </summary>
+    public OwnedContractName InitName { get; init; }
+    /// <summary>
+    /// A list of events generated by a smart contract.
+    /// </summary>
+    public IList<ContractEvent> Events { get; init; }
+
+    internal ContractInitializedEvent(Concordium.Grpc.V2.ContractInitializedEvent initializedEvent)
```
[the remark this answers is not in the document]
schwartz-concordium: When I look at it why are we using`From`? To me it feels like we are copying Rust style.

src/Types/Address.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// Should only be called after TypeCase check.  
+    /// </summary>
+    public ContractAddress ContractAddress { get; init; }
+    /// <summary>
+    /// Specify if this is a Account or Contract address. Should always be called first
+    /// to determine address type.
+    /// </summary>
+    public AddressTypeCase TypeCase { get; init; }
+    
+    internal Address(Grpc.V2.Address address)
```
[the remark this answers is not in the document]
schwartz-concordium: Why are we using `From`?

src/Types/TransactionStatus.cs
schwartz-concordium: Yes one has to match at runtime - I’m quite confident this is standard C#.

src/Types/RejectReason.cs
schwartz-concordium: Not sure I’m follow - can you give an example?

src/Client/ConcordiumClient.cs
```diff
@@ -94,14 +94,14 @@ SignedAccountTransaction<T> transaction
     /// case, then the sequence number is reliable.
     /// </returns>
     /// <exception cref="RpcException">The call failed.</exception>
-    public (Types.AccountSequenceNumber, bool) GetNextAccountSequenceNumber(
-        Types.AccountAddress accountAddress
+    public (AccountSequenceNumber, bool) GetNextAccountSequenceNumber(
+        AccountAddress accountAddress
```
abizjak, line 98: Generally don't do this, i.e., make many unrelated changes in many places.

PRs should be as minimal as reasonable, and we should not be making multiple independent changes everywhere.

src/Types/ContractInitializedEvent.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <summary>
+    /// Represents a contract address.
+    /// </summary>
+    public ContractAddress ContractAddress { get; init; }
+    /// <summary>
+    /// Represents a CCD amount.
+    /// </summary>
+    public CcdAmount Amount { get; init; }
+    /// <summary>
+    /// A contract name.
```
abizjak, line 27: Please elaborate that this will be in the format `init_*`, i.e., there will be this `init_` prefix.

src/Types/HashBytes.cs
```diff
@@ -0,0 +1,13 @@
+using Google.Protobuf;
+
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// A general wrapper of a 32-size hash. This is used to add type safety to
+/// a hash that is used in different context.
+/// </summary>
+public record HashBytes : Hash
```
abizjak, line 9: I'm not sure I understand how this adds type safety. In the rust world we use a phantom type parameter for this, but here there are no type parameters. So how does this add safety?

src/Types/OwnedParameter.cs
```diff
@@ -0,0 +1,6 @@
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// Parameter to the init function or entrypoint. Owned version.
```
abizjak, line 4: I think this "Owned" prefix does not make sense in C#.

It makes sense in Rust since there is a difference between Vec<u8> and &[u8], but in C# you in any case have this implicit indirection, so you can't really have &[u8] equivalent.

So I recommend removing all these Owned prefixes.

src/Types/OwnedReceiveName.cs
```diff
@@ -0,0 +1,9 @@
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// A receive name (owned version). Expected format:
+/// "<contract_name>.<func_name>". Most methods are available only on the
+/// [`ReceiveName`] type, the intention is to access those via the
+/// [`as_receive_name`](OwnedReceiveName::as_receive_name) method.
+/// </summary>
+public record struct OwnedReceiveName(string OwnedReceive);
```
abizjak, line 9: Also here.

src/Types/BlockItemSummary.cs
```diff
@@ -0,0 +1,368 @@
...
+    public bool IsReject() => this.TryGetRejectedAccountTransaction(out _);
+
+    /// <summary>
+    /// When transaction is rejected this method will return true and rejected reason
+    /// will be return in out param.
+    /// </summary>
+    /// <param name="rejectReason">Output with rejected reason if rejected.</param>
+    /// <returns>bool representing if transaction is rejected.</returns>
+    /// <exception cref="MissingEnumException{DetailsOneofCase}">Throws exception when returned type not known</exception>
+    public bool TryGetRejectedAccountTransaction(out RejectReason? rejectReason)
```
abizjak, line 50: Is this the way to do things in C#? I would have thought returning an Option type would be the way to go, rather than out parameters.

src/Types/BlockItemSummary.cs
```diff
@@ -0,0 +1,368 @@
...
+        };
+    }
+
+    /// <summary>
+    /// Returns true and set sender account if of account transaction type.
+    /// </summary>
+    /// <param name="sender">Output given when sender present.</param>
+    /// <returns>If sender account present.</returns>
+    /// <exception cref="MissingEnumException{DetailsOneofCase}">Throws exception when returned type not known</exception>
+    public bool TryGetSenderAccount(out AccountAddress? sender)
```
abizjak, line 72: Same here. Is it not more ergonomic to return values seeing that the caller does not have to create the variable first?

schwartz-concordium: AccountAddress is nullable and hence the caller would not need to create a variable.

src/Types/BlockItemSummary.cs
schwartz-concordium: Yes - there are some language extension packages which includes options and those functional language operators but it isn’t very C# alike.

I wrapped my head around this a few times when I wrote the code and I’m still convinced this is the best C# approach. Try… patterns is well known in the C# world.

src/Types/AccountAddress.cs
```diff
@@ -155,7 +155,16 @@ public AccountAddress GetNthAlias(uint n)
     public override bool Equals(object? obj) =>
         obj is not null && obj is AccountAddress other && this.Equals(other);
 
-    public override int GetHashCode() => this.GetBytes().GetHashCode();
+    public override int GetHashCode()
```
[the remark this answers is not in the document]
schwartz-concordium: The former one is a bug - the hash is taken on the reference pointer of the byte array. Hence the value would be different.
I have already written this in Viktor’s PR and asked him to update these all places where IEquatable interface is overwritten. I needed it here since I used the struct in a HashSet.

src/Client/ConcordiumClient.cs
schwartz-concordium: I was told to run `dotnet format` such that code aligned with our editerconfig settings - this is why these changes are present.

src/Types/BlockItemSummary.cs
abizjak: I don't understand. To get anything out they need to do something like
```
var address;
if TryGetSenderAccount(out address) {
    ...
}
```

I had assumed there would at least be a standard Optional<T> type that would make this cleaner. BUt if there isn't then I suppose we must live with this C-style stuff.

schwartz-concordium: I believe this is the way to do it - somehow one needs to make a check. I can give a review with Ali and see if he knows a better pattern.

To me it seems  however quite much the same between C# and Rust

C#

```
if (TryGetSenderAccount(out var address) {
   // do something with address
}
```

Rust 

```
if let Some(address) = senderAccount() {
   // do something with address
}
```

please let me know if I’m missing something - it always fun fun to learn :-)

abizjak: I did not know you can declare variables inside parameters to a function call.

Now that I do know I hate it since it adds complexity. :P 

But OK.

Also note in Rust you would probably do
```

```
let Some(address) = senderAccount() else {
    // Handle failure.
};
// use address here
```
to avoid too much nesting.

Comment (schwartz-concordium) [reply src/Types/BlockItemSummary.cs:72]: I didn’t know you could do the else afterwards - nice since nesting is exactly one of the things in Rust I find difficult and make it difficult to read :-)

File: ConcordiumNetSdk.sln (hunk 6 above)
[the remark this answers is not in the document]
Comment (schwartz-concordium) [reply ConcordiumNetSdk.sln]: Change back since the value changing it is this PR is to small.

File: ConcordiumNetSdk.sln (hunk 8 above)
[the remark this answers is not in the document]
Comment (schwartz-concordium) [reply ConcordiumNetSdk.sln]: Changed back

File: ConcordiumNetSdk.sln (hunk 9 above)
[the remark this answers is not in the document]
Comment (schwartz-concordium) [reply ConcordiumNetSdk.sln]: Changed back

File: ConcordiumNetSdk.sln [hunk 21]
```diff
@@ -39,39 +43,53 @@ Global
...
+		{8D147631-91A8-451D-BB4C-02E01FC01B2E}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{8D147631-91A8-451D-BB4C-02E01FC01B2E}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {D9EB56EA-568E-4763-9538-FA327D35D016}
 	EndGlobalSection
+	GlobalSection(NestedProjects) = preSolution
```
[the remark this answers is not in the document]
schwartz-concordium: Changed back

examples/Common/Example.cs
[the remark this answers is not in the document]
schwartz-concordium: Changed back

examples/Common/ExampleOptions.cs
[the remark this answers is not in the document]
schwartz-concordium: Changed back

ConcordiumNetSdk.sln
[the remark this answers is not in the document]
schwartz-concordium: Changed back

examples/Common/ExampleOptionsWithKeys.cs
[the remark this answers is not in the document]
schwartz-concordium: Changed back

examples/GetAccountInfo/RawClient.GetAccountInfo.csproj
```diff
@@ -5,10 +5,11 @@
     <TargetFramework>net6.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <RootNamespace>RawClient.GetAccountInfo</RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\Common\Concordium.Sdk.Examples.Common.csproj" />
```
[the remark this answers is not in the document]
schwartz-concordium: Changed back

examples/GetBlockItemStatus/Program.cs
```diff
@@ -0,0 +1,113 @@
...
+internal sealed class GetBlockItemSummaryOptions
+{
+    [Option(HelpText = "Transaction hash to lookup", Required = true)]
+    public string TransactionHash { get; set; }
+
+    [Option(HelpText = "URL representing the endpoint where the gRPC V2 API is served.", Required = true)]
+    public string Endpoint { get; set; }
+}
+
+internal static class ExampleHelpers
```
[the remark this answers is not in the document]
schwartz-concordium: I made the Parse helper method more clean and avoid the Try pattern

examples/GetBlockItemStatus/Program.cs
```diff
@@ -0,0 +1,113 @@
...
+public static class Program
+{
+    /// <summary>
+    /// An example showing how one can query transaction status from a node.
+    /// </summary>
+    /// <param name="args">GetBlockItemSummaryOptions
+    /// Example: --endpoint http://node.testnet.concordium.com/ --transactionhash 143ca4183d0bb204000ad08e0fd5792985c808861b97f3b81cb9016ad39d09d2</param>
+    public static async Task Main(string[] args)
+    {
+        if (!ExampleHelpers.TryParse(args, out GetBlockItemSummaryOptions? options))
```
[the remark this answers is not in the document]
schwartz-concordium: Resolve this and keep track of discussion in above comments

examples/Examples/Transactions/GetBlockItemStatus/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+        var transactionHash = TransactionHash.From(options.TransactionHash);
+        var port = options.Port;
+
+        using var client = new ConcordiumClient(endpoint, port);
+
+        Console.WriteLine("Query node...");
+        var transactionStatus = await client.GetBlockItemStatusAsync(transactionHash);
+
+        // Conditional on the transaction status different actions can be handled.
+        switch (transactionStatus)
```
ali-concordium, line 59: Consider using switch expression for consistency, as it's applied elsewhere in the PR:

In example:
_ = transactionStatus switch
{
    TransactionStatusReceived received => HandleTransactionStatusReceived(received),
    TransactionStatusFinalized finalized => HandleTransactionStatusFinalized(finalized),
    TransactionStatusCommitted committed => HandleTransactionStatusCommitted(committed),
    _ => throw new NotImplementedException(transactionStatus.GetType().ToString())
};

examples/Examples/Transactions/GetBlockItemStatus/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+        var endpoint = new Uri(options!.Endpoint);
+        var transactionHash = TransactionHash.From(options.TransactionHash);
+        var port = options.Port;
+
+        using var client = new ConcordiumClient(endpoint, port);
+
+        Console.WriteLine("Query node...");
+        var transactionStatus = await client.GetBlockItemStatusAsync(transactionHash);
+
+        // Conditional on the transaction status different actions can be handled.
```
ali-concordium, line 58: The comment doesn't provide any value, as the code is pretty explicit and self-explanatory. Hence I think it should be removed. Alternatively, you could encapsulate the switch in a method named "HandleTransactionStatus" and pass the transactionStatus variable to it, but I recommend to just leave it as is.

ali-concordium (approved):

src/Types/HashBytes.cs
schwartz-concordium: Made types explicit

examples/Examples/Transactions/GetBlockItemStatus/Program.cs
schwartz-concordium: I unfortunatly can’t do this since the functions are void. The compilers complains with

`Only assignment, call, increment, decrement, await expression, and new object expressions can be used as a statement`

src/Types/ContractInitializedEvent.cs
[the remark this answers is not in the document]
vikt0r0: Some subscribe to not having logic and throwing exceptions in constructors, and at least that's what I did to the largest possible extent in the codebase. I think it would be nice if we were consistent.

As for the reasoning for my choice of that, `From` does checks that throw exceptions, and while you would expect the parameters of the constructors to be stored, this assumption goes away when using `From`, so it provides for some further encapsulation in general.

src/Types/RejectReason.cs
vikt0r0: So currently, we need to map each enum variant to a field declared in this class, no?

So if we had something like `InvalidEncryptedAmountTransferProof : RejectReason<InvalidEncryptedAmountTransferProof>`, then the `ReasonCase` field could be factored into the abstract `RejectReason<T>` class and we would not have to use this mapping to determine the non-null fields, since `InvalidEncryptedAmountTransferProof` would have only the fields specific to it?

src/Types/Address.cs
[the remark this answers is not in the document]
vikt0r0: See my other comment.

src/Types/BlockItemSummary.cs
schwartz-concordium: I have had a talk with Ali and this is the best approach as we see it.

vikt0r0 (approved):

src/Types/BlockHash.cs
```diff
@@ -27,6 +28,12 @@ private BlockHash(string blockHashAsBase16String)
     private BlockHash(byte[] blockHashAsBytes)
         : base(blockHashAsBytes) { }
 
+    /// <summary>
+    /// Creates an instance from a byte string object given from generated code. Only internal use.
+    /// </summary>
+    /// <param name="byteString">A block hash represented as a immutable array.</param>
+    internal static BlockHash From(ByteString byteString) => new(byteString.ToByteArray());
```
abizjak, line 35: Shouldn't this be checking length? Or be stated as a precondition that it is unsafe.

abizjak (approved):
Looks ok.

Please address outstanding comments before merging.

src/Types/BlockHash.cs
schwartz-concordium: This is done in the parent constructor

vikt0r0:
> @schwartz-concordium unfortunately Viktor squashed all the commits which means there are now conflicts.
> 
> Can you please get this resolved.

Sorry about that. Let me know if I should do anything. I can revert the squashed merge commit and restore the branch if needed.

PR #47, opened 2023-05-26
Fix example links

Bargsteen:
## Purpose

Fix example links in README (and thus also in the generated docs).
I am not sure whether we _need_ the nested example folders (`./examples/Examples/..`), but this at least fixes the links.

## Changes

- Change links

## Checklist

      hard-to-understand areas.


vikt0r0 (approved):
Nice 🤘🏼

Bargsteen:
Do we need to republish the docs or is that handled automatically?

PR #56, opened 2023-07-25
Made to lower invariant and fixed minor bugs after gRPC update

schwartz-concordium:
## Purpose

Made sure all call which creates Hex string make them to lower invariants.

This since from gRPC v1 we have from the nodes returned hex strings in lower invariants. Hence if some have stored these references in some storage and selecting using case sensitive queries, then their logic will be broken.

Changelog not updated since all changes in this PR only affects code which hasn’t been released yet.

## Changes

Make Hex strings consistent with earlier protocol.

Also fix minor bugs after gRPC update.

## Checklist

- [X] The code compiles without warnings.
      hard-to-understand areas.
- [-] (If necessary) I have updated the CHANGELOG.

### DIFF
--- a/src/Types/BakerInfo.cs
+++ b/src/Types/BakerInfo.cs
@@ -1,0 +1,31 @@
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// Information about a baker.
+/// </summary>
+/// <param name="BakerId">
+/// Identity of the baker. This is actually the account index of
+/// the account controlling the baker.
+/// </param>
+/// <param name="BakerElectionVerifyKey">
+/// Baker's public key used to check whether they won the lottery or not.
+/// </param>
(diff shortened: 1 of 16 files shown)


src/Types/BakerInfo.cs
```diff
@@ -0,0 +1,31 @@
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// Information about a baker.
+/// </summary>
+/// <param name="BakerId">
+/// Identity of the baker. This is actually the account index of
```
bisgardo, line 7: Nit: "actually" sounds unnecessarily surprised to me. Would just write something like "Account index of the account associated with the baker." It being the ID of the baker is implied in the name.

bisgardo (approved):
LGTM. Ideally all the changes should have been covered by unit tests.

PR #58, opened 2023-07-28
Updated SDK to align with Protocol 6

schwartz-concordium:
## Purpose

Update SDK to align with new consensus protocol and hence changes done in gRPC API
https://github.com/Concordium/concordium-grpc-api/commit/13e23942735a427b188b6a27909044ff6493f50a.

This PR at the same time prepare for release 4.0.0.

## Changes

  - `ConsensusInfo`
    - `SlotDuration` is now an nullable field, only present in protocols 1-5.
    - a new field `ConcordiumBftDetails` is added, that is present if protocol
      version is 6 or higher
    - Bugfix: `BlockLastArrivedTime` was wrongly mapped from `BlockLastReceivedTime`.
  - `BlockInfo`
    - `BlockSlot` is nullable, and only present in protocols 1-5
    - new fields `Round` and `Epoch` that are present in protocol 6 or higher.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Types/ConcordiumBftDetails.cs
+++ b/src/Types/ConcordiumBftDetails.cs
@@ -1,0 +1,33 @@
+using Concordium.Sdk.Helpers;
+
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// Parameters pertaining to the Concordium BFT consensus.
+/// </summary>
+/// <param name="CurrentTimeoutDuration">The current duration to wait before a round times out.</param>
+/// <param name="Round">The current round.</param>
+/// <param name="Epoch">The current epoch.</param>
+/// <param name="TriggerBlockTime">
+/// The first block in the epoch with timestamp at least this is considered
+/// to be the trigger block for the epoch transition.
+/// </param>
+public sealed record ConcordiumBftDetails(TimeSpan CurrentTimeoutDuration, Round CurrentRound, Epoch Epoch, DateTimeOffset TriggerBlockTime)
+{
+    internal static bool TryFrom(Grpc.V2.ConsensusInfo info, out ConcordiumBftDetails? details)
+    {
+        if (info.CurrentTimeoutDuration == null || info.CurrentRound == null ||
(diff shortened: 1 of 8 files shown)


abizjak (approved):
Looks good.

abizjak:
Also an emoji :rocket: to fullfil the challenge.

src/Types/ConcordiumBftDetails.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// Parameters pertaining to the Concordium BFT consensus.
+/// </summary>
+/// <param name="CurrentTimeoutDuration">The current duration to wait before a round times out.</param>
+/// <param name="Round">The current round.</param>
+/// <param name="Epoch">The current epoch.</param>
+/// <param name="TriggerBlockTime">
+/// The first block in the epoch with timestamp at least this is considered
```
ali-concordium, line 12: Consider rephrasing to:

_The first block in the epoc with **a** timestamp **equal to or later than this timestamp,** is considered ...._

src/Types/ConsensusInfo.cs
```diff
@@ -129,12 +136,15 @@ public sealed record ConsensusInfo(
...
+    double? FinalizationPeriodEmsd,
+    ConcordiumBftDetails? ConcordiumBftDetails
     )
 {
-    internal static ConsensusInfo From(Grpc.V2.ConsensusInfo consensusInfo) =>
-        new
-        (
+    internal static ConsensusInfo From(Grpc.V2.ConsensusInfo consensusInfo)
+    {
+        var _ = ConcordiumBftDetails.TryFrom(consensusInfo, out var concordiumBftDetails);
```
ali-concordium, line 145: Why do we implement the try method, if we discard the return value and pass concordiumBftDetails regardless of its value?

ali-concordium (approved):

ali-concordium (approved):

PR #61, opened 2023-08-10
Bug fix found under CCD Scan Protocol 6 update regression test 

schwartz-concordium:
## Purpose

Bugfix found when doing regression change on CCD Scan due to protocol 6 update.

## Changes

* https://concordium.atlassian.net/browse/CBW-1230
   `TransactionCount` in `BlockInfo` had wrong mapping and used `TransactionsSize` from node.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/tests/IntegrationTests/Client/GetBlockInfo.cs
+++ b/tests/IntegrationTests/Client/GetBlockInfo.cs
@@ -13,6 +13,20 @@ public sealed class GetBlockInfo : Tests
     }
 
     [Fact]
+    public async Task GivenBlockWithOneTransaction_WhenGetBlockInfo_ThenTransactionCountOne()
+    {
+        // Arrange
+        var blochHashString = this.GetString("blockHashOneTransaction");
+
+
+        // Act
+        var info = await this.Client.GetBlockInfoAsync(new Given(BlockHash.From(blochHashString)));
+
+        // Assert
+        info.Response.TransactionCount.Should().Be(1);
+    }
+
+    [Fact]
     public async Task GivenAbsoluteHeightAboveBlockHeight_WhenCallGetBlockInfo_ThenThrowException()
     {
         // Arrange
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 4.0.1
+- Bugfix
+  - `TransactionCount` in `BlockInfo` had wrong mapping and used `TransactionsSize` from node.
+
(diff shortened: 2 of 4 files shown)


tests/IntegrationTests/Client/GetBlockInfo.cs
```diff
@@ -12,6 +12,20 @@ public GetBlockInfo(ITestOutputHelper output) : base(output)
     {
     }
 
+    [Fact]
+    public async Task GivenBlockWithOneTransaction_WhenGetBlockInfo_ThenTransactionCountOne()
+    {
+        // Arrange
+        var blochHashString = this.GetString("blockHashOneTransaction");
```
ali-concordium, line 19: If you have some extra time, look into AutoFixture.
It can save a lot of boilerplate testing code when arranging dummy values, mocks and objects.

ali-concordium: We should also add coverage for when a block contains more than one transaction & verify the expected behavior when the hash doesn't point to valid block. Meaning Info cannot be found.

limemloh: Please don't introduce `AutoFixture`, it hides the test data which most of the time will be crucial for the test.

tests/IntegrationTests/Client/GetBlockInfo.cs
```diff
@@ -12,6 +12,20 @@ public GetBlockInfo(ITestOutputHelper output) : base(output)
...
+
+
+        // Act
+        var info = await this.Client.GetBlockInfoAsync(new Given(BlockHash.From(blochHashString)));
```
limemloh, line 23: I don't understand the point of this test, I would expect this to fail when parsing the blockhash from string and even if this continues I would expect the block to not have any transactions.
What am I missing?

limemloh (changes requested):

tests/IntegrationTests/Client/GetBlockInfo.cs
schwartz-concordium: Removed since none of the other SDK’s have test like this.

tests/IntegrationTests/Client/GetBlockInfo.cs
schwartz-concordium: The block hash was loaded from configuration and was the block has of a block with one transaction.

Has been removed since none of the other SDK’s have test like this.

limemloh (approved):

PR #64, opened 2023-08-16
Add missing documentation when project is packaged.

schwartz-concordium:
## Purpose

Documentation was missing when adding library as nuget packages and hovering over methods and classes.

## Changes

Added documentation when project is packaged.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium.Sdk.csproj
+++ b/src/Concordium.Sdk.csproj
@@ -21,6 +21,9 @@
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageLicenseExpression>MPL-2.0</PackageLicenseExpression>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <DocumentationFile>bin\Release\net6.0\Concordium.Sdk.xml</DocumentationFile>
+  </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


src/Concordium.Sdk.csproj
```diff
@@ -21,6 +21,9 @@
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageLicenseExpression>MPL-2.0</PackageLicenseExpression>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <DocumentationFile>bin\Release\net6.0\Concordium.Sdk.xml</DocumentationFile>
+  </PropertyGroup>
```
limemloh, line 26: I think we should use [`GenerateDocumentationFile`](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#generatedocumentationfile) instead, and build this for any build, since it will show the warnings related to documentation.
```suggestion
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
  </PropertyGroup>
```

limemloh (approved):
Good thing, that you caught this.
Please add an issue with the task of fixing the 106 warnings from documentation.

PR #77, opened 2023-11-03
Add name helpers

schwartz-concordium:
## Purpose

Add helpers to get contract name and entrypoint.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

- [ ] I accept the above linked CLA.

### DIFF
--- a/src/Types/ContractName.cs
+++ b/src/Types/ContractName.cs
@@ -40,7 +40,7 @@ public sealed record ContractName
     /// Get the contract name part of <see cref="Name"/>.
     /// </summary>
     /// <returns>Contract name</returns>
-    public string GetContractName() => this.Name[(this.Name.IndexOf('_')+1)..];
+    public string GetContractName() => this.Name[(this.Name.IndexOf('_') + 1)..];
 
     /// <summary>
     /// Validation error of contract name.
--- a/src/Types/ContractName.cs
+++ b/src/Types/ContractName.cs
@@ -39,8 +39,8 @@ public sealed record ContractName
     /// <summary>
     /// Get the contract name part of <see cref="Name"/>.
     /// </summary>
-    /// <returns>Contract name</returns>
-    public string GetContractName() => this.Name[(this.Name.IndexOf('_') + 1)..];
+    /// <returns>Contract identification name</returns>
+    public ContractIdentifier GetContractName() => new(this.Name[(this.Name.IndexOf('_') + 1)..]);
 
     /// <summary>
     /// Validation error of contract name.
--- a/src/Types/ContractName.cs
+++ b/src/Types/ContractName.cs
@@ -37,6 +37,12 @@ public sealed record ContractName
     }
 
     /// <summary>
+    /// Get the contract name part of <see cref="Name"/>.
+    /// </summary>
+    /// <returns>Contract name</returns>
+    public string GetContractName() => this.Name[(this.Name.IndexOf('_')+1)..];
+
+    /// <summary>
     /// Validation error of contract name.
     /// </summary>
     public enum ValidationError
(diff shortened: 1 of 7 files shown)


src/Types/ReceiveName.cs
```diff
@@ -34,6 +34,18 @@ public static bool TryParse(string name, out (ReceiveName? ReceiveName, Validati
...
+    /// Get the contract name part of <see cref="Receive"/>.
+    /// </summary>
+    /// <returns>Contract name</returns>
+    public string GetContractName() => this.Receive[..this.Receive.IndexOf('.')];
+
+    /// <summary>
+    /// Get entrypoint part of <see cref="Receive"/> which is the entrypoint called on the contract.
+    /// </summary>
+    /// <returns>Entrypoint</returns>
+    public string GetEntrypoint() => this.Receive[(this.Receive.IndexOf('.') + 1)..];
```
limemloh, line 47: I think we should have an `Entrypoint` type, instead of using string.
We will need this when we add support for contract update transactions.

src/Types/ReceiveName.cs
```diff
@@ -34,6 +34,18 @@ public static bool TryParse(string name, out (ReceiveName? ReceiveName, Validati
         return validate;
     }
 
+    /// <summary>
+    /// Get the contract name part of <see cref="Receive"/>.
+    /// </summary>
+    /// <returns>Contract name</returns>
+    public string GetContractName() => this.Receive[..this.Receive.IndexOf('.')];
```
limemloh, line 41: I think this should return a `ContractName` instead of a string.

src/Types/ContractName.cs
```diff
@@ -36,6 +36,12 @@ public static bool TryParse(string name, out (ContractName? ContractName, Valida
         return validate;
     }
 
+    /// <summary>
+    /// Get the contract name part of <see cref="Name"/>.
+    /// </summary>
+    /// <returns>Contract name</returns>
+    public string GetContractName() => this.Name[(this.Name.IndexOf('_') + 1)..];
```
limemloh, line 43: This should probably be named differently, since it won't behave like the `GetContractName` method on `ReceveName` (after addressing my comment there).

src/Types/ReceiveName.cs
schwartz-concordium: Unfortunately we already have a type called `ContractName`. It follows the same naming conventions as in base library
https://github.com/Concordium/concordium-base/blob/a3a16407c82ad49de404b3968ba8507f86314681/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs#L1180
This type does however contain both prefix `init_` and the contract name.

Looking at it now aren't the naming wrong in both SDKs? Contract name should only be the part without the `init_` prefix as far as I understand.

I see two options
* Make a new type called something like `ContractNamePart`
* Change name of old type from `ContractName` to `InitName` and make a major bump

limemloh: > Looking at it now aren't the naming wrong in both SDKs? Contract name should only be the part without the init_ prefix as far as I understand.

Yes, that is how it is done in the JS SDK, but not necessarily the best solution.

> I see two options
> 
> - Make a new type called something like `ContractNamePart`
> - Change name of old type from `ContractName to `InitName` and make a major bump

Since `ContractName` and `InitName` are 1 to 1, and the only difference is how it is represented as a string, I think we should try to make the current `ContractName` represent both without breaking changes. It will simplify the API since there won't be two types for essentially the same information.

schwartz-concordium: Resolved through slack. We introduce a new type, since for a lot of use cases the user only have the contract name part. Ex. when working with contract- updated and interrupted events.

limemloh (approved):

Issue #68, opened 2023-08-23
Using schema to support smart contract JSON parameters


limemloh:
**Task description**

Add functionality for using the smart contract schema to convert bytes to and from JSON. 
This is relevant for smart contract parameters, events and return types.

**Sub-tasks**
- [x] Extract schema bytes from a smart contract module source ([embedding](https://developer.concordium.software/en/mainnet/smart-contracts/general/contract-schema.html#embedding-schemas-on-chain)).
- [x] Functionality for turning bytes and the schema into JSON.
- [ ] Functionality for turning JSON and the schema into bytes.

limemloh:
For the last two items, we will need to decide whether to implement and maintain all of this functionality directly in C# or whether to reuse the Rust implementation through FFI.
The FFI will be needed anyway at a later point for encrypted transfers.

abizjak:
As we discussed at the daily today, unless there are blocking issues we should not implement this in C#.

There is just too much complexity to implement and maintain.

limemloh:
Remaining tasks are tracked in #94

PR #78, opened 2023-11-03
Add interop functionality

schwartz-concordium:
## Purpose

Added interop functionality to deserialize schema, entrypoint messages and events.

Covers first two subtask of
https://github.com/Concordium/concordium-net-sdk/issues/68

https://concordium.atlassian.net/browse/CBW-1508

## Changes

- Added functionality to deserialization from module schema
  - Contract entrypoint messages
  - Contract events
  - Module schema

## Checklist

      hard-to-understand areas.


schwartz-concordium:
Validate build order between C# and Rust before merge.

src/Types/ContractEvent.cs
```diff
@@ -9,4 +12,31 @@ public sealed record ContractEvent(byte[] Bytes)
...
     public string ToHexString() => Convert.ToHexString(this.Bytes).ToLowerInvariant();
+
+    /// <summary>
+    /// Deserialize event from <see cref="schema"/>.
+    /// </summary>
+    /// <param name="schema">Module schema in hexadecimal.</param>
+    /// <param name="contractName">Contract name.</param>
+    /// <returns><see cref="Bytes"/> deserialized.</returns>
+    /// <exception cref="InteropBindingException">Thrown when event wasn't able to be deserialized from schema.</exception>
+    public string? GetDeserializeEvent(
```
limemloh, line 23: **Suggestion:** Make signature like `TryParse` and indicate in the name that this can fail (`Try` prefix).

Since this is parsing the bytes in the event, it should probably follow the same style as `TryParse`, if you would like we can also provide a function which throws.

src/Types/VersionedModuleSchema.cs
```diff
@@ -0,0 +1,38 @@
...
+
+/// <summary>
+/// Module schema embedded within <see cref="VersionedModuleSource"/>.
+/// </summary>
+public sealed record VersionedModuleSchema
+{
+    /// <summary>
+    /// Module schema in hexadecimal.
+    /// </summary>
+    public string Schema { get; }
```
limemloh, line 11: **Suggestion:** Store the schema as raw bytes instead of hex encoding.

src/Types/ContractEvent.cs
limemloh, line 23: **Suggestion:** The output should indicate that this is JSON instead of just `string`, maybe [`Utf8JsonReader`](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/use-utf8jsonreader) is a good match, since it won't actually deserialize the JSON string and the user can bring their own class for this.

src/Types/ContractTraceElement.cs
```diff
@@ -71,7 +73,55 @@ public sealed record Updated(
...
+            schema.Schema,
+            this.ReceiveName.GetContractName(),
+            this.ReceiveName.GetEntrypoint(),
+            this.Message.ToHexString(),
+            schema.Version);
+
+    /// <summary>
+    /// Deserialize message from <see cref="schema"/>.
+    /// </summary>
+    /// <param name="schema">Module schema in hexadecimal.</param>
```
limemloh, line 95:
```suggestion
    /// <param name="schema">Module schema.</param>
```

src/Interop/InteropBinding.cs
```diff
@@ -0,0 +1,157 @@
...
+    [DllImport(DllName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "get_event_contract")]
+    private static extern bool GetEventContract(string schema, FfiByteOption schema_version, string contract_name, string value, ref IntPtr result);
+
+    /// <summary>
+    /// Get module schema in a human interpretable form.
+    /// </summary>
+    /// <param name="schema">Module schema in hexadecimal</param>
+    /// <param name="schemaVersion">Optional schema version if present from module</param>
+    /// <returns>Module schema in a human interpretable form</returns>
+    internal static string? SchemaDisplay(string schema, ModuleSchemaVersion? schemaVersion)
```
limemloh, line 32: When is the return value `null`? Please add this to the documentation.

limemloh (changes requested):

src/Concordium.Sdk.csproj
```diff
@@ -68,6 +73,38 @@
...
+    <Content Include="..\rust-bindings\Cargo.toml">
+      <Link>rust-bindings\Cargo.toml</Link>
+    </Content>
+    <Content Include="..\rust-bindings\src\lib.rs">
+      <Link>rust-bindings\src\lib.rs</Link>
+    </Content>
+    <Content Include="..\rust-bindings\test-data\cis2_wCCD_sub">
+      <Link>rust-bindings\test-data\cis2_wCCD_sub</Link>
+    </Content>
+  </ItemGroup>
```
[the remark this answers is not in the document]
schwartz-concordium: Mistake

src/Types/ContractEvent.cs
```diff
@@ -9,4 +12,31 @@ public sealed record ContractEvent(byte[] Bytes)
...
+        VersionedModuleSchema schema,
+        string contractName
```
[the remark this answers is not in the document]
schwartz-concordium: Should be able to use just from a Contract Update event, where you don’t want to map a string from “contract_name.entrypoint” -> “contract_name” -> “init_contract_name” -> “contract_name”. The last two steps are just unnecessary.

.github/workflows/build-test-format.yml
```diff
@@ -7,22 +7,52 @@ on:
...
+    - name: Set up Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: ${{ env.RUST_VERSION }}
+        override: true
+        components: rustfmt, clippy
+
+    - name: Format Rust Project
+      run: cargo fmt --manifest-path=./rust-bindings/Cargo.toml --check
```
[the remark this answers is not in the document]
abizjak: Yes, and please use the same nightly formatter version for that.

rust-bindings/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+/// * 'result' - Parsed schema if the call succeeded or the error message in case of failure.
+///
+/// # Returns
+///
+/// If the call succeeded or not.
+///
+/// # Safety
+///
+/// This function is marked as unsafe because it performs operations that are not checked by the
+/// Rust compiler.
```
abizjak, line 46: Usually the safety section should explain how to use it safely.

.github/workflows/build-test-format.yml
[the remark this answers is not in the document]
schwartz-concordium: I have added the `rustfmt.toml`. 


In `concordium-base` pipelines format and clippy runs in two different jobs with two different rust version. May I ask why we do this?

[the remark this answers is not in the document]
abizjak: The stable formatter does not support the unstable options we use. THe main reason here is that rustfmt very very slowly stabilizes anything. The options we are using have been there for years now, but are not stabilized yet.

src/Types/ContractEvent.cs
schwartz-concordium: Then penalty of throwing the exception is already present since it is thrown from the interops operation. I have had this discussion with some from UFT team when introducing alike in the CCD scan, and we concluded best C# practice was to have an exception thrown. 
I believe in must cases the user would like to have the exception if it fails, instead of a silent swallow with a `Try`.

src/Interop/InteropBinding.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            FreeIfNonzero(result);
+        }
+    }
+
+    private static void FreeIfNonzero(IntPtr ptr)
+    {
+        if (ptr != IntPtr.Zero)
+        {
+            Marshal.FreeHGlobal(ptr);
```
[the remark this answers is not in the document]
schwartz-concordium: have made a callback approach

src/Types/ContractEvent.cs
schwartz-concordium: Sum up of discussion p2p. It would be nice to show output is json but since `Utf8JsonReader` is a `ref struct` it is limited how it can be used. We discussed `JsonNode` or `JsonDocument` but in the end it will be a overhead.

We will write it clearly in the documentation and then let it be up to the user to parse the result as efficient as they like.

src/Interop/InteropBinding.cs
schwartz-concordium: Changed return type to not be null

src/Concordium.Sdk.csproj
```diff
@@ -68,6 +73,38 @@
     </Protobuf>
   </ItemGroup>
 
+
+  <ItemGroup>
+    <!-- Include librust.dll on Windows -->
+    <None Include="../rust-bindings/target/release/librust_bindings.dll" Condition="'$(IsWindows)'=='true'">
```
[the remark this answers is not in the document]
schwartz-concordium: Sum up discussion p2p. We will need to have all binaries built for multiple platforms. We will look into making a `nuspec` file which helps us package correctly.

I have made an independent PR which will fix this.
https://github.com/Concordium/concordium-net-sdk/pull/79

.github/workflows/build-test-format.yml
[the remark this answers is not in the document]
schwartz-concordium: I have updated the pipeline and added format and clippy to multiple steps. Please verify and resolve if ok.

rust-bindings/src/lib.rs
schwartz-concordium: Sections updated

.github/workflows/build-test-format.yml
```diff
@@ -7,22 +7,70 @@ on:
...
+  RUST_CLIPPY: 1.68
+
 jobs:
-  build-test-format:
+  rustfmt:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
```
abizjak, line 20:
```suggestion
    - uses: actions/checkout@v4
```

.github/workflows/build-test-format.yml
```diff
@@ -7,22 +7,70 @@ on:
...
+    - name: Set up Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: ${{ env.RUST_FMT }}
+        override: true
+        components: rustfmt
+
+    - name: Format Rust Project
+      run: cargo fmt --manifest-path=./rust-bindings/Cargo.toml --all --check
```
abizjak, line 33:
```suggestion
    - name: Format Rust Project
      run: |
        rustup default ${{ env.RUST_FMT }}
        rustup component add rustfmt
        cargo fmt --manifest-path=./rust-bindings/Cargo.toml --all --check
```

I recommend avoiding actions-rs in new pipelines since it is unmaintained. I recently migrated all of concordium-base jobs away from using these actions. The base ubuntu image has everything we need already.

.github/workflows/build-test-format.yml
```diff
@@ -7,22 +7,70 @@ on:
...
       with:
         submodules: recursive
+
+    - name: Set up Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: ${{ env.RUST_CLIPPY }}
+        override: true
+        components: clippy
```
abizjak, line 50:
```suggestion
      uses: actions-rs/toolchain@v1
      run: |
        rustup default ${{ env.RUST_CLIPPY }}
        rustup component add clippy
```

rust-bindings/src/lib.rs
```diff
@@ -0,0 +1,310 @@
+use std::{ffi::CStr, os::raw::c_char};
+
```
abizjak, line 2:
```suggestion
```

And let the formatter sort them.

src/Concordium.Sdk.csproj
```diff
@@ -45,29 +49,50 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.Linq.Async" Version="6.0.1" />
+    <PackageReference Include="WebAssembly" Version="1.2.1" />
```
abizjak, line 52: Can you explain what this is for?

src/Concordium.Sdk.csproj
```diff
@@ -45,29 +49,50 @@
...
   </ItemGroup>
 
+
+  <ItemGroup>
+    <!-- Include librust.dll on Windows -->
+    <None Include="../rust-bindings/target/release/librust_bindings.dll" Condition="'$(IsWindows)'=='true'">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <!-- Include librust.dylib on macOS -->
+    <None Include="../rust-bindings/target/release/librust_bindings.dylib" Condition="'$(IsOSX)'=='true'">
```
abizjak, line 83: Is this x86 or aarch64? Will it depend on what machine this thing is built?

abizjak (commented):
I'm not a fan of including all of these binary files (modules, schemas) with no documentation on how they were generated, and how they should be updated. Experience suggests this is not maintainable.

So please include that.

If I missed the documentation please point me to it.

src/Concordium.Sdk.csproj
schwartz-concordium: We are using it to parse the wasm module and “entering” the custom sections to extract the schema.

src/Concordium.Sdk.csproj
schwartz-concordium: Yes dependent on which platform it is build on the `librust_bindings.*` will be different.

Please note that this is required for local execution. The packaging, which includes all binaries for Linux, iOS, and Windows, will be handled in another PR.

Regarding your query about 'Is this x86 or aarch64,' could you provide further clarification or context? I'm unsure about the specific information you're seeking.

src/Types/ContractTraceElement.cs
```diff
@@ -71,7 +72,52 @@ public sealed record Updated(
...
+    /// </summary>
+    /// <param name="schema">Module schema.</param>
+    /// <returns>List of deserialized json uft8 encoded events. Possible null if this was returned from deserialization.</returns>
+    /// <exception cref="InteropBindingException">Thrown if an event wasn't able to be deserialized from schema.</exception>
+    public IList<Utf8Json> GetDeserializedEvents(VersionedModuleSchema schema)
+    {
+        var deserialized = new List<Utf8Json>(this.Events.Count);
+        foreach (var contractEvent in this.Events)
+        {
+            var deserializeEvent = contractEvent.GetDeserializeEvent(schema, this.ReceiveName.GetContractName());
```
limemloh, line 114: **Minor thing:** Since `GetContractName` is allocating a new string, it should probably be moved out of the loop.

src/Types/RejectReason.cs
```diff
@@ -182,11 +182,22 @@ public sealed record OutOfEnergy : IRejectReason;
...
 public sealed record RejectedReceive(int RejectReason, ContractAddress ContractAddress, ReceiveName ReceiveName,
-    Parameter Parameter) : IRejectReason;
+    Parameter Parameter) : IRejectReason
+{
+    /// <summary>
+    /// Deserialize message from <see cref="schema"/>.
+    /// </summary>
+    /// <param name="schema">Versioned module schema.</param>
+    /// <returns><see cref="Parameter"/> deserialized as json uft8 encoded.</returns>
+    /// <exception cref="InteropBindingException">Thrown when message wasn't able to be deserialized form schema.</exception>
```
limemloh, line 197:
```suggestion
    /// <exception cref="InteropBindingException">Thrown when message wasn't able to be deserialized using the schema.</exception>
```

tests/UnitTests/Data/module.schema_embedded.wasm.hex
limemloh, line 1: **Note (non-blocking):** You seem to switch a bit between using hex encoded and binary files for the test data.
Maybe just have all of them in binary?

limemloh (approved):
This looks good to me.

I would prefer not to merge this, until we have the packaging PR done (or nearly done), since we otherwise won't be able to release after this PR is merged.

README.md
```diff
@@ -41,6 +41,18 @@ in your project root. It can also be used as a GIT submodule by embedding the cl
 git clone https://github.com/Concordium/concordium-net-sdk --recurse-submodules
 ```
 
+## Rust bindings
+
+The .NET SDK uses Foreign Function Interface (FFI) functions to code written in Rust. The Rust code is located in `./rust-bindings`.
+
+Whenever the .NET SDK is build the Rust code is autimatically compiled. This is done as a `PreBuildEvent` defined in the `Concordium.Sdk.csproj` file. Hence any changes
```
abizjak, line 48:
```suggestion
Whenever the .NET SDK is build the Rust code is automatically compiled. This is done as a `PreBuildEvent` defined in the `Concordium.Sdk.csproj` file. Hence any changes
```

rust-bindings/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+) -> Result<Vec<u8>> {
+    let module_schema = VersionedModuleSchema::new(schema, &schema_version)?;
+    let parameter_type = module_schema.get_receive_param_schema(contract_name, entrypoint)?;
+    let deserialized = deserialize_type_value(value, &parameter_type, true)?;
+    Ok(deserialized)
+}
+
+unsafe fn slice_from_ptr<'a, T>(data: *const T, size: usize) -> &'a [T] {
+    std::slice::from_raw_parts(data, size)
+}
```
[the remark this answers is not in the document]
abizjak: I agree.

src/Concordium.Sdk.csproj
abizjak: There are currently two mac platforms widely used, the older x86-64 intel macs and the newer M{1,2,3} based ones. You can use the x86 code on m1 macs via rosetta but not the other way around.

So my concern is that we package the SDK in the right way so that it's actually usable on both types of platform.

src/Types/ModuleSchemaVersion.cs
```diff
@@ -0,0 +1,28 @@
+namespace Concordium.Sdk.Types;
+
+/// <summary>
+/// Represents the different module schema versions.
+/// </summary>
+public enum ModuleSchemaVersion
+{
+    /// <summary>
+    /// Version not known and possible embedded in schema.
```
abizjak, line 9:
```suggestion
    /// Version not known and possibly embedded in schema.
```

abizjak:
> > I'm not a fan of including all of these binary files (modules, schemas) with no documentation on how they were generated, and how they should be updated. Experience suggests this is not maintainable.
> > So please include that.
> > If I missed the documentation please point me to it.
> 
> I have added a section in the `README.md` about the rust code.
> 
> Are you also referring to test data?

Thanks for including the documentation on the relationship with the Rust code. But I was also talking about all those schema files and modules.

They should be documented on how they are to be updated and what is important about them.

abizjak (approved):
Please include a description of how the test files are meant to be maintained.

src/Concordium.Sdk.csproj
schwartz-concordium: In this PR, which handles packaging, I have made a pipeline which builds on mac os. 
https://github.com/Concordium/concordium-net-sdk/pull/79

Looking at the mac os build step I can see the version used for rust is `1.68-x86_64-apple-darwin`. So from what I understand we are packaging correctly.
https://github.com/Concordium/concordium-net-sdk/actions/runs/6971280381/job/18970984682

I've followed the flow used in the Java SDK release pipeline. If we haven’t encountered any issues related to the Java release I wouldn’t expect any on the .NET.
https://github.com/Concordium/concordium-java-sdk/blob/main/.github/workflows/release.yml#L114

tests/UnitTests/Data/module.schema_embedded.wasm.hex
schwartz-concordium: Yes that is a good point. I will continue adding only binaries and let those generated in hexadecimal be there. I will write in documentation that one should prefer generating binary test data.

schwartz-concordium:
> > > I'm not a fan of including all of these binary files (modules, schemas) with no documentation on how they were generated, and how they should be updated. Experience suggests this is not maintainable.
> > > So please include that.
> > > If I missed the documentation please point me to it.
> > 
> > 
> > I have added a section in the `README.md` about the rust code.
> > Are you also referring to test data?
> 
> Thanks for including the documentation on the relationship with the Rust code. But I was also talking about all those schema files and modules.
> 
> They should be documented on how they are to be updated and what is important about them.

I have added a section now.

src/Concordium.Sdk.csproj
abizjak: The Java flow hasn't really been tested, it's very new.

But it does look that macos-latest is still intel based so it is fine.

Thanks for checking up.

PR #80, opened 2023-11-28
Get block items

rasmus-kirk:
## Purpose

_Describe the purpose of the pull request, link to issue describing the problem, etc.

## Changes

_Describe the changes that were needed.

## Checklist

- [ ] My code follows the style of this project.
- [ ] I have performed a self-review of the changes.
      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

- [ ] I accept the above linked CLA.

### DIFF
--- a/src/Helpers/Deserialization.cs
+++ b/src/Helpers/Deserialization.cs
@@ -1,0 +1,47 @@
+using System.Buffers.Binary;
+
+namespace Concordium.Sdk.Helpers;
+
+/// <summary>
+/// Error on deserialization.
+/// </summary>
+public enum DeserialErr
+{
+    TooShort,
+	InvalidModuleVersion,
+}
+
+/// <summary>
+/// Helpers for deserializing data.
+/// </summary>
+public static class Deserial
+{
+    /// <summary>
+    /// Creates a uint from a byte array.
+    /// </summary>
+    public static bool TryDeserialU32(byte[] input, int offset, out (uint? Uint, DeserialErr? Error) output)
+    {
+		var offset_input = input.Skip(offset).ToArray();
+
+        if (offset_input.Length < 4) {
+			output = (null, DeserialErr.TooShort);
+			return false;
+		}
+
+		var bytes = offset_input.Take(4).ToArray();
+
+        output = (BinaryPrimitives.ReadUInt32BigEndian(bytes), null);
+		return true;
+    }
+
+	private static void PrintBytes(String msg, byte[] bytes) {
+		Console.WriteLine(msg);
+		foreach (byte b in bytes) {
+			Console.Write(b);
+			Console.Write(" ");
+		}
+		Console.Write("\n");
+	} 
+}
+
+
--- a/src/Helpers/Deserialization.cs
+++ b/src/Helpers/Deserialization.cs
@@ -3,71 +3,81 @@ using System.Buffers.Binary;
 namespace Concordium.Sdk.Helpers;
 
 /// <summary>
-/// Error on deserialization.
-/// </summary>
-public enum DeserialErr
-{
-    TooShort,
-	InvalidModuleVersion,
-	InvalidTransactionType,
-	InternalError,
-	InvalidLength,
-}
-
-
-
-/// <summary>
 /// Helpers for deserializing data.
 /// </summary>
 public static class Deserial
 {
     /// <summary>
-    /// Creates a uint from a byte array.
+    /// Creates a ushort from a byte array.
     /// </summary>
-    public static bool TryDeserialU32(byte[] input, int offset, out (uint? Uint, String? Error) output)
+    public static bool TryDeserialU16(byte[] input, int offset, out (ushort? Uint, string? Error) output)
     {
-        if (input.Length < 4) {
-			var msg = $"Invalid length in TryDeserialU32. Must be longer than 4, but was {input.Length}";
-			output = (null, msg);
-			return false;
-		}
+        if (input.Length < sizeof(ushort))
+        {
+            var msg = $"Invalid length in TryDeserialU32. Must be longer than {sizeof(ushort)}, but was {input.Length}";
+            output = (null, msg);
+            return false;
+        }
 
-		var offset_input = input.Skip(offset).ToArray();
+        var offset_input = input.Skip(offset).ToArray();
 
-		var bytes = offset_input.Take(4).ToArray();
+        var bytes = offset_input.Take(sizeof(ushort)).ToArray();
 
-        output = (BinaryPrimitives.ReadUInt32BigEndian(bytes), null);
-		return true;
+        output = (BinaryPrimitives.ReadUInt16BigEndian(bytes), null);
+        return true;
     }
 
     /// <summary>
     /// Creates a uint from a byte array.
     /// </summary>
-    public static bool TryDeserialU64(byte[] input, int offset, out (ulong? Ulong, String? Error) output)
+    public static bool TryDeserialU32(byte[] input, int offset, out (uint? Uint, string? Error) output)
     {
-        if (input.Length < 8) {
-			var msg = $"Invalid length in TryDeserialU32. Must be longer than 4, but was {input.Length}";
-			output = (null, msg);
-			return false;
-		}
+        if (input.Length < sizeof(uint))
+        {
+            var msg = $"Invalid length in TryDeserialU32. Must be longer than 4, but was {input.Length}";
+            output = (null, msg);
+            return false;
+        }
 
-		var offset_input = input.Skip(offset).ToArray();
+        var offset_input = input.Skip(offset).ToArray();
 
-		var bytes = offset_input.Take(8).ToArray();
+        var bytes = offset_input.Take(sizeof(uint)).ToArray();
 
-        output = (BinaryPrimitives.ReadUInt64BigEndian(bytes), null);
-		return true;
+        output = (BinaryPrimitives.ReadUInt32BigEndian(bytes), null);
+        return true;
     }
 
-	// TODO: Debug tool remove
-	private static void PrintBytes(String msg, byte[] bytes) {
-		Console.WriteLine(msg);
-		foreach (byte b in bytes) {
-			Console.Write(b);
-			Console.Write(" ");
-		}
-		Console.Write("\n");
-	} 
+    /// <summary>
+    /// Creates a uint from a byte array.
+    /// </summary>
+    public static bool TryDeserialU64(byte[] input, int offset, out (ulong? Ulong, string? Error) output)
+    {
+        if (input.Length < sizeof(ulong))
+        {
+            var msg = $"Invalid length in TryDeserialU32. Must be longer than {sizeof(ulong)}, but was {input.Length}";
+            output = (null, msg);
+            return false;
+        }
+
+        var offset_input = input.Skip(offset).ToArray();
+
+        var bytes = offset_input.Take(sizeof(ulong)).ToArray();
+
+        output = (BinaryPrimitives.ReadUInt64BigEndian(bytes), null);
+        return true;
+    }
+
+    // TODO: Debug tool remove
+    public static void PrintBytes(string msg, byte[] bytes)
+    {
+        Console.WriteLine(msg);
+        foreach (var b in bytes)
+        {
+            Console.Write(b);
+            Console.Write(" ");
+        }
+        Console.Write("\n");
+    }
 }
 
 
--- a/src/Helpers/Deserialization.cs
+++ b/src/Helpers/Deserialization.cs
@@ -9,6 +9,8 @@ public enum DeserialErr
 {
     TooShort,
 	InvalidModuleVersion,
+	InvalidTransactionType,
+	InternalError,
 }
 
 /// <summary>
@@ -34,6 +36,7 @@ public static class Deserial
 		return true;
     }
 
+	// TODO: Debug tool remove
 	private static void PrintBytes(String msg, byte[] bytes) {
 		Console.WriteLine(msg);
 		foreach (byte b in bytes) {
--- a/src/Helpers/Deserialization.cs
+++ b/src/Helpers/Deserialization.cs
@@ -11,7 +11,10 @@ public enum DeserialErr
 	InvalidModuleVersion,
 	InvalidTransactionType,
 	InternalError,
+	InvalidLength,
 }
+
+
 
 /// <summary>
 /// Helpers for deserializing data.
@@ -21,18 +24,38 @@ public static class Deserial
     /// <summary>
     /// Creates a uint from a byte array.
     /// </summary>
-    public static bool TryDeserialU32(byte[] input, int offset, out (uint? Uint, DeserialErr? Error) output)
+    public static bool TryDeserialU32(byte[] input, int offset, out (uint? Uint, String? Error) output)
     {
-		var offset_input = input.Skip(offset).ToArray();
-
-        if (offset_input.Length < 4) {
-			output = (null, DeserialErr.TooShort);
+        if (input.Length < 4) {
+			var msg = $"Invalid length in TryDeserialU32. Must be longer than 4, but was {input.Length}";
+			output = (null, msg);
 			return false;
 		}
+
+		var offset_input = input.Skip(offset).ToArray();
 
 		var bytes = offset_input.Take(4).ToArray();
 
         output = (BinaryPrimitives.ReadUInt32BigEndian(bytes), null);
+		return true;
+    }
+
+    /// <summary>
+    /// Creates a uint from a byte array.
+    /// </summary>
+    public static bool TryDeserialU64(byte[] input, int offset, out (ulong? Ulong, String? Error) output)
+    {
+        if (input.Length < 8) {
+			var msg = $"Invalid length in TryDeserialU32. Must be longer than 4, but was {input.Length}";
+			output = (null, msg);
+			return false;
+		}
+
+		var offset_input = input.Skip(offset).ToArray();
+
+		var bytes = offset_input.Take(8).ToArray();
+
+        output = (BinaryPrimitives.ReadUInt64BigEndian(bytes), null);
 		return true;
     }
 
--- a/src/Transactions/AccountSignatureMap.cs
+++ b/src/Transactions/AccountSignatureMap.cs
@@ -66,3 +66,55 @@ public sealed record AccountSignatureMap
         return AccountSignatureMap.Create(dict);
     }
 }
+
+
+/// <summary>Index of a key in an authorizations update payload.</summary>
+/// <param name="Value">The key index represented by a <see cref="byte"/>.</param>
+public sealed record UpdateKeysIndex(byte Value);
+
+/// <summary>A map from <see cref="UpdateKeysIndex"/> to signatures.</summary>
+public sealed record UpdateInstructionSignatureMap
+{
+    /// <summary>Internal representation of the map.</summary>
+    public ImmutableDictionary<UpdateKeysIndex, byte[]> Signatures { get; init; }
+
+    /// <summary>Initializes a new instance of the <see cref="UpdateInstructionSignatureMap"/> class.</summary>
+    /// <param name="signatures">A map from account key indices to signatures.</param>
+    private UpdateInstructionSignatureMap(Dictionary<UpdateKeysIndex, byte[]> signatures) => this.Signatures = signatures.ToImmutableDictionary();
+
+    /// <summary>Creates a new instance of the <see cref="UpdateInstructionSignatureMap"/> class.</summary>
+    /// <param name="signatures">A map from account key indices to signatures.</param>
+    /// <exception cref="ArgumentException">A signature is not 64 bytes.</exception>
+    public static UpdateInstructionSignatureMap Create(Dictionary<UpdateKeysIndex, byte[]> signatures)
+    {
+        // Signatures are 64-byte ed25519 signatures and therefore 64 bytes.
+        if (signatures.Values.Any(signature => signature.Length != 64))
+        {
+            throw new ArgumentException($"Signature should be {64} bytes.");
+        }
+        return new UpdateInstructionSignatureMap(signatures);
+    }
+
+    /// <summary>Converts the account signature map to its corresponding protocol buffer message instance.</summary>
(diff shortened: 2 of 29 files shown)


src/Transactions/DeployModule.cs
```diff
@@ -0,0 +1,84 @@
...
+            return false;
+        };
+        if (bytes[0] != TransactionType)
+        {
+            var msg = $"Invalid transaction type in `DeployModule.TryDeserial`. expected {TransactionType}, found {bytes[0]}";
+            output = (null, msg);
+            return false;
+        }
+
+        output = (new DeployModule(module.VersionedModuleSource), null);
```
rasmus-kirk, line 78: This warns that `module.VersionedModuleSource` might be `null`, because the compiler doesn't know that it never will be by construction of the previous functions.

Is this an actual issue?

src/Transactions/TransferWithMemo.cs
```diff
@@ -28,17 +49,72 @@ private static byte[] Serialize(CcdAmount amount, AccountAddress receiver, OnCha
...
+        {
+            var msg = $"Invalid transaction type in `Transfer.TryDeserial`. expected {TransactionType}, found {bytes[0]}";
+            output = (null, msg);
+            return false;
+        };
+
+        var trxTypeLength = sizeof(TransactionType);
+        var accountLength = (int)AccountAddress.BytesLength;
+        var amountLength = (int)CcdAmount.BytesLength;
+        var memoLength = bytes.Length - trxTypeLength - accountLength - amountLength;
```
limemloh, line 86: **Request change:** This is relying on the provided bytes to only contain exactly the payload bytes, which we don't really require anywhere else. 

The length of the memo is encoded in the payload, using UInt16BE:
See https://github.com/Concordium/concordium-base/blob/main/haskell-src/Concordium/Types.hs#L807.

limemloh (changes requested):

src/Transactions/BlockItem.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    internal static BlockItem From(Grpc.V2.BlockItem blockItem) =>
+        new(
+            TransactionHash.From(blockItem.Hash.ToString()),
+            blockItem.BlockItemCase switch
+            {
+                BlockItemCase.AccountTransaction => SignedAccountTransaction.From(blockItem.AccountTransaction),
+                BlockItemCase.CredentialDeployment => CredentialDeployment.From(blockItem.CredentialDeployment),
+                BlockItemCase.UpdateInstruction => UpdateInstruction.From(blockItem.UpdateInstruction),
+                BlockItemCase.None => throw new NotImplementedException(),
```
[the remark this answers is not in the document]
limemloh: No, just let the current wild card take care of it.
So essentially just remove the line with `BlockItemCase.None`

src/Transactions/TransferWithMemo.cs
```diff
@@ -18,6 +20,36 @@ public sealed record TransferWithMemo(CcdAmount Amount, AccountAddress Receiver,
...
+    /// This should reflect the transaction-specific costs defined here:
+    /// https://github.com/Concordium/concordium-base/blob/78f557b8b8c94773a25e4f86a1a92bc323ea2e3d/haskell-src/Concordium/Cost.hs
+    /// </summary>
+    private readonly EnergyAmount _transactionCost = new(300);
+
+    /// <summary>
+    /// Gets the size (number of bytes) of the payload.
+    /// </summary>
+    internal override PayloadSize Size() => new(
+        this.Memo.Length() +
```
schwartz-concordium, line 48: Here `this.Memo.Length()` is used and below `OnChainData.MaxLength`.

Is that correct?

src/Transactions/TransferWithMemo.cs
```diff
@@ -28,17 +60,70 @@ private static byte[] Serialize(CcdAmount amount, AccountAddress receiver, OnCha
...
+        var minSize = sizeof(TransactionType) + AccountAddress.BytesLength + CcdAmount.BytesLength + sizeof(ushort);
+        if (bytes.Length < minSize)
+        {
+            var msg = $"Invalid length in `TransferWithMemo.TryDeserial`. Expected at least {minSize}, found {bytes.Length}";
+            output = (null, msg);
+            return false;
+        };
+        if (bytes[0] != TransactionType)
+        {
+            var msg = $"Invalid transaction type in `Transfer.TryDeserial`. expected {TransactionType}, found {bytes[0]}";
```
schwartz-concordium, line 89:
```suggestion
            var msg = $"Invalid transaction type in `Transfer.TryDeserial`. Expected {TransactionType}, found {bytes[0]}”;
```

src/Transactions/TransferWithMemo.cs
```diff
@@ -28,17 +60,70 @@ private static byte[] Serialize(CcdAmount amount, AccountAddress receiver, OnCha
...
+        var amountBytes = bytes[(trxTypeLength + accountLength + memoLength)..];
+        if (!CcdAmount.TryDeserial(amountBytes, out var amount))
+        {
+            output = (null, amount.Error);
+            return false;
+        };
+
+        if (amount.Amount == null || account.AccountAddress == null || memo.OnChainData == null)
+        {
+            throw new DeserialInvalidResultException();
```
schwartz-concordium, line 121: No exceptions in try pattern

src/Transactions/UpdateInstruction.cs
```diff
@@ -0,0 +1,66 @@
...
+/// Update sequence numbers are per update type and the minimum value is 1.
+/// </summary>
+public record UpdateSequenceNumber(ulong SequenceNumber)
+{
+    internal static UpdateSequenceNumber From(Grpc.V2.UpdateSequenceNumber sequenceNumber) =>
+        new(sequenceNumber.Value);
+}
+
+/// <summary>The payload for an UpdateInstruction.</summary>
+public interface IUpdateInstructionPayload { }
```
schwartz-concordium, line 63: I would make a interface if there is only one inheritance and that structure is a wrapper around bytes.

src/Types/OnChainData.cs
```diff
@@ -25,6 +29,11 @@ public sealed record OnChainData : IEquatable<OnChainData>
     /// <param name="bytes">Data represented by at most <see cref="MaxLength"/> bytes.</param>
     private OnChainData(byte[] bytes) => this._value = bytes;
 
+    /// <summary>
+    /// Gets the length (number of bytes) of the data.
+    /// </summary>
+    internal uint Length() => (uint)this._value.Length;
```
schwartz-concordium, line 35: It appears that we're continuously switching between different casts of uint and int. Wouldn’t it be better to just store as int and then only cast when interacting with the chain?

src/Types/VersionedModuleSource.cs
```diff
@@ -1,11 +1,10 @@
+using System.Buffers.Binary;
 using Concordium.Sdk.Exceptions;
+using Concordium.Sdk.Helpers;
 using WebAssembly;
 
 namespace Concordium.Sdk.Types;
 
-/// <summary>
-/// Contains source code of a versioned module where inherited classes are concrete versions.
```
schwartz-concordium, line 7: Was this header wrong?

src/Types/VersionedModuleSource.cs
```diff
@@ -100,9 +164,46 @@ versionedModuleSource.ModuleCase switch
...
+    public static ModuleV0 From(byte[] source)
+    {
+        if (source.Length > MaxLength)
+        {
+            throw new ArgumentException(
+                $"Size of a data is not allowed to exceed {MaxLength} bytes."
+            );
+        }
+
+        return new ModuleV0(source.ToArray());
```
schwartz-concordium, line 186: Do we need to copy the array?

src/Types/OnChainData.cs
rasmus-kirk: I also dislike it. I made it uint as the `BytesLength` in CcdAmount, Transfer, AccountAddress, etc. are all of uint. So this is to keep consistency.

src/Types/VersionedModuleSource.cs
rasmus-kirk: No this is definitely a merge error, my bad.

src/Types/VersionedModuleSource.cs
rasmus-kirk: I would think so? Otherwise, mutating the provided array would mutate the module. That doesn't seem intuitive to me.

src/Transactions/AccountTransactionPayload.cs
```diff
@@ -12,39 +14,107 @@ namespace Concordium.Sdk.Transactions;
...
+            case TransactionType.TransferWithSchedule:
+            case TransactionType.UpdateCredentials:
+            case TransactionType.EncryptedAmountTransferWithMemo:
+            case TransactionType.TransferWithScheduleAndMemo:
+            case TransactionType.ConfigureBaker:
+            case TransactionType.ConfigureDelegation:
+                parsedPayload = (new RawPayload(payload.ToArray()), null);
+                break;
+            default:
+                throw new MissingEnumException<TransactionType>((TransactionType)payload.First());
```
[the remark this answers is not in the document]
rasmus-kirk: I talked to Emil Holm, about this and the conclusion was no. I'll poke at it just to confirm though.

src/Transactions/DeployModule.cs
```diff
@@ -0,0 +1,95 @@
...
+        if (bytes.Length < minLength)
+        {
+            var msg = $"Invalid input length in `DeployModule.TryDeserial`. expected at least {minLength}, found {bytes.Length}";
+            output = (null, msg);
+            return false;
+        }
+
+        if (bytes[0] != TransactionType)
+        {
+            var msg = $"Invalid transaction type in `DeployModule.TryDeserial`. expected {TransactionType}, found {bytes[0]}";
```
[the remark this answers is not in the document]
rasmus-kirk: This is a problem in several places, thank you. I'll grep/sed it.

src/Transactions/RegisterData.cs
```diff
@@ -31,5 +56,42 @@ private static byte[] Serialize(OnChainData data)
...
+        var memoBytes = bytes[sizeof(TransactionType)..];
+        if (!OnChainData.TryDeserial(memoBytes, out var memo))
+        {
+            output = (null, memo.Error);
+            return false;
+        };
+
+        if (memo.OnChainData == null)
+        {
+            throw new DeserialInvalidResultException();
```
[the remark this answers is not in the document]
rasmus-kirk: See my other comment.

src/Transactions/Transfer.cs
```diff
@@ -16,24 +17,93 @@ public sealed record Transfer(CcdAmount Amount, AccountAddress Receiver) : Accou
...
+        var amountBytes = bytes[((int)AccountAddress.BytesLength + sizeof(TransactionType))..];
+        if (!CcdAmount.TryDeserial(amountBytes, out var amount))
+        {
+            output = (null, amount.Error);
+            return false;
+        };
+
+        if (amount.Amount == null || account.AccountAddress == null)
+        {
+            throw new DeserialInvalidResultException();
```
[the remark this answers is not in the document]
rasmus-kirk: See my other comment.

src/Transactions/TransferWithMemo.cs
rasmus-kirk: See my other comment.

src/Transactions/AccountTransactionPayload.cs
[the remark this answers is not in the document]
rasmus-kirk: I can confirm, `(TransactionType)100` does not throw an exception, and will also be caught by the `MissingEnumException`.

src/Transactions/BlockItem.cs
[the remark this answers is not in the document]
rasmus-kirk: So far I have removed the `None` case, so it should be caught by the `MissingEnumException`. I personally think that's fine since even if it does happen, it would be a bug in the node as far as I know. I don't think we need explicit errors for that.

I am open to discussion though.

src/Transactions/TransferWithMemo.cs
rasmus-kirk: I removed the `Serialize` methods since they were private and kind of clunky. Now the `MemoryStream` just uses `Size().Size` instead, so it's defined in a single place.

rasmus-kirk: I removed the `Serialize` methods since they were private anyways. This lets us define the MemoryStream directly from the `Size` and solves this inconsistency.

src/Types/OnChainData.cs
rasmus-kirk: What do you think is the best solution here?

src/Transactions/BlockItem.cs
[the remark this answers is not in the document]
rasmus-kirk: Wait the `None` case is demanded by the formatter I'll add it back with a `MissingEnumException`...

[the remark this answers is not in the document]
limemloh: @rasmus-kirk You are right in the `None` case should never happen and reaching the case means something went wrong in the node. I think it would be good to introduce an exception representing this, just as @schwartz-concordium suggest. Maybe something like `UnexpectedNodeResponse`.

src/Transactions/AccountSignatureMap.cs
```diff
@@ -56,4 +56,66 @@ public Grpc.V2.AccountSignatureMap ToProto()
...
+        return Create(dict);
+    }
+}
+
+
+/// <summary>Index of a key in an authorizations update payload.</summary>
+public sealed record UpdateKeysIndex(byte Value);
+
+/// <summary>A map from <see cref="UpdateKeysIndex"/> to signatures.</summary>
+public sealed record UpdateInstructionSignatureMap
```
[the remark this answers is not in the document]
limemloh: I am not sure, I understand the issue here, it is already a `class` (`record class`), and to my knowledge this is only changing the behavior of equality.
Also, I don't see this type being mutated and is only used for reading from the chain, meaning immutability is not an issue.

Similarly, for the user will encounter `AccountSignatureMap` when reading from chain, and probably only be constructed internally in the SDK and not by the user.

src/Transactions/AccountTransactionPayload.cs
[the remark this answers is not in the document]
limemloh: Yes, casting an integer to an enum cannot cause an exception, but the value will instead just keep the integer representation https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum#conversions.

However, `payload.First()` will throw an exception if the payload is empty, but this would mean something went wrong in the node.

src/Transactions/TransferWithMemo.cs
```diff
@@ -18,27 +20,101 @@ public sealed record TransferWithMemo(CcdAmount Amount, AccountAddress Receiver,
...
+            var msg = $"Invalid transaction type in `Transfer.TryDeserial`. Expected {TransactionType}, found {bytes[0]}";
+            output = (null, msg);
+            return false;
+        };
+
+        var trxTypeLength = sizeof(TransactionType);
+        var accountLength = (int)AccountAddress.BytesLength;
+        var memoLength = BinaryPrimitives.ReadUInt16BigEndian(bytes[(trxTypeLength + accountLength)..]) + sizeof(ushort);
+
+        var accountBytes = bytes[trxTypeLength..];
```
limemloh, line 78: **Suggestion:** I have been trying to figure out what range indexing actually does for `Span`, but without luck, so just to be sure that we don't copy here, we should use the `.Slice` instead.

```suggestion
        var accountBytes = bytes.Slice(trxTypeLength);
```
Likewise for the other places using `bytes[x..]` in this function

src/Types/VersionedModuleSource.cs
```diff
@@ -14,6 +16,17 @@ public abstract record VersionedModuleSource
     public byte[] Source { get; }
     private readonly Lazy<Module> _module;
 
+    internal const uint MaxLength = 8 * 65536;
```
limemloh, line 19: **Request change:** The max size in bytes is 65535, so I don't think we should multiply by 8 here. Also it seems to be off by one.
```suggestion
    internal const uint MaxLength = 65535;
```

src/Types/VersionedModuleSource.cs
```diff
@@ -92,6 +126,41 @@ versionedModuleSource.ModuleCase switch
...
+    public static bool TryDeserial(ReadOnlySpan<byte> bytes, out (VersionedModuleSource? VersionedModuleSource, string? Error) output)
+    {
+        if (bytes.Length < VersionedModuleSource.MinSerializedLength)
+        {
+            output = (null, $"The given byte array in `VersionModuleSourceFactory.TryDeserial`, is too short. Must be longer than {VersionedModuleSource.MinSerializedLength}.");
+            return false;
+        }
+
+        var version = BinaryPrimitives.ReadUInt32BigEndian(bytes);
+        var length = BinaryPrimitives.ReadUInt32BigEndian(bytes[sizeof(int)..]);
```
limemloh, line 144: **Suggestion:** Again, we should have a comment here to why this is "safe" from exceptions

src/Types/VersionedModuleSource.cs
limemloh: I don't expect users will mutate these, and avoiding the copy would be good.
Maybe just document that this assumes the bytes are never mutated, otherwise the user will need to copy before calling.

Also, since it is an array, they can only change the value of the bytes and not add or remove anything, meaning the above check of the length is still kept valid in the case they mutate.

src/Types/VersionedModuleSource.cs
```diff
@@ -100,9 +169,46 @@ versionedModuleSource.ModuleCase switch
...
+    /// <param name="hexString">The WASM-module represented as a hex encoded string representing at most "MaxLength" bytes.</param>
+    /// <exception cref="ArgumentException">The supplied string is not a hex encoded WASM-module representing at most "MaxLength" bytes.</exception>
+    public static ModuleV0 FromHex(string hexString)
+    {
+        try
+        {
+            var value = Convert.FromHexString(hexString);
+            return From(value);
+        }
+        catch (Exception e)
```
limemloh, line 206: **Request change:** This catches the invalid length exception thrown by `From` and still provide the exception below about hex encoding.
Maybe you can just skip the try catch and let the exception from `Convert.FromHexString` pop further up.

src/Transactions/TransferWithMemo.cs
schwartz-concordium: `Span` is just a contiguous region in memory so any slice would reference the same. Try
```
    [Fact]
    public void SpanSlicing()
    {
        // Arrange
        var someArray = new byte[] { 1, 2, 3, 4, 5, 6 };
        var someSpan = someArray.AsSpan();

        var someSlice = someSpan[..3];

        // Act
        someSlice[2] = 9; // Change value in slice affect initial array

        // Assert
        Assert.Equal(9, someArray[2]);
    }
```

Also if you use `Slice` the intellisense suggest you should use the range indexer.

src/Helpers/Deserialization.cs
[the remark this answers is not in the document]
rasmus-kirk: Yes

src/Transactions/TransferWithMemo.cs
rasmus-kirk: Actually, I have been using `Slice` but the formatter will change it so using `Slice` would give issues with it. I'd rather keep the range indexer.

src/Types/VersionedModuleSource.cs
rasmus-kirk: I think this was supposed to be `2*sizeof(ushort) + 65535` for the versioning. But maybe I'm misunderstanding the `MaxLength` here. I'll check some documentation/rust code.

src/Transactions/AccountSignatureMap.cs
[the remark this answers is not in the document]
limemloh: Having read a bit more about records, I think they are the right thing here, since the type is just holding data and does not provide any (interesting) logic. The generated `toString` function is useful as well. I suggest that we fix the behavior of equality instead.

Changing this would also make it inconsistent with the nearly identical type `AccountSignatureMap`.

src/Transactions/TransferWithMemo.cs
limemloh: Great! Then ignore this suggestion.

src/Types/VersionedModuleSource.cs
rasmus-kirk: No that's wrong, I got the `8 * 65536` from here:
https://github.com/Concordium/concordium-base/blob/6d9ad3500d0150062d8596715f6920ff33678c89/rust-src/concordium_base/src/constants.rs#L19

limemloh: My mistake, I was using the max size of the parameter, not the module. So you were correct all along `512 kB = 8 * 65536`.

src/Transactions/AccountSignatureMap.cs
```diff
@@ -56,4 +56,110 @@ public Grpc.V2.AccountSignatureMap ToProto()
...
+    {
+        // Based on https://stackoverflow.com/questions/1646807/quick-and-simple-hash-code-combinations
+        unchecked
+        {
+            var hash = 17;
+            foreach (var (key, val) in this.Signatures)
+            {
+                hash = (hash * 31) + key.GetHashCode();
+                hash = (hash * 31) + Helpers.HashCode.GetHashCodeByteArray(val);
+            }
```
limemloh, line 87: **Issue:** `Signatures` is an unordered collection, and I am not sure if we can rely on the ordering being the same, for two equal signatures.
We need this hash code to be independent of the iteration ordering.

src/Transactions/AccountSignatureMap.cs
```diff
@@ -56,4 +56,110 @@ public Grpc.V2.AccountSignatureMap ToProto()
...
```
limemloh, line 161: **Issue:** `Signatures` is an unordered collection, and I am not sure if we can rely on the ordering being the same, for two equal signatures.
We need this hash code to be independent of the iteration ordering.

PR #86, opened 2023-12-27
Change InteropBindingException to public

schwartz-concordium:
## Purpose

Change InteropBindingException such that it can be caught by users.

### DIFF
--- a/src/Exceptions/InteropBindingException.cs
+++ b/src/Exceptions/InteropBindingException.cs
@@ -5,7 +5,7 @@ namespace Concordium.Sdk.Exceptions;
 /// <summary>
 /// Thrown when a interop call failed with possible error as message.
 /// </summary>
-internal sealed class InteropBindingException : Exception
+public sealed class InteropBindingException : Exception
 {
     private const string EmptyErrorMessage = "Empty error message returned";
 


src/Exceptions/InteropBindingException.cs
```diff
@@ -5,7 +5,7 @@ namespace Concordium.Sdk.Exceptions;
 /// <summary>
 /// Thrown when a interop call failed with possible error as message.
 /// </summary>
-internal sealed class InteropBindingException : Exception
+public sealed class InteropBindingException : Exception
```
limemloh, line 8: I am not a big fan of leaking the internals of the SDK, but if you need this we can merge it

limemloh (approved):

src/Exceptions/InteropBindingException.cs
abizjak: Why would the users ever want to catch this? What can they do in that case? Is this also used for normal error handling?

I think that a design should rather be talking about user meaningful errors and exceptions, not about the fact that there is some FFI binding somewhere. Users should not need to know that or care about it.

limemloh: This is used for the errors returned by the FFI to rust, so they are needed by the user. 
I have discussed this with Søren and the plan is now to rename and change this exception to contain an error code (enum matching the rust side error) and the error message provided by rust.

schwartz-concordium:
Closing in favour of PR https://github.com/Concordium/concordium-net-sdk/pull/93

Issue #48, opened 2023-05-26
Implement smart contract transactions.


vikt0r0:
**Task description**

Add support for deploy module, initialize contract and update contract transactions. This does not include functionality for parameter serialization and deserialization with schemas, but rather just uses binary representations.

**Implement**
- [X] Deploy module.
- [x] Initialize contract. #89 
- [x] Update contract. #91

PR #91, opened 2024-01-25
Add ContractUpdate transaction payload

limemloh:
## Purpose

Add basic implementation for sending update contract transaction.
Part of #48.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 ## Unreleased changes
   - Added
     - New GRPC-endpoint: `GetBlockItems`
-    - New transaction `DeployModule`
+    - New transactions: `DeployModule` and `UpdateContract`.
     - The function `Prepare` has been removed from the `AccountTransactionPayload` class, but is implemented for all subclasses except `RawPayload`.
-    - Added serialization and deserialization for all instances of `AccountTransactionPayload`
+    - Added serialization and deserialization for all instances of `AccountTransactionPayload`.
 
   - Breaking
     - The function `GetTransactionSpecificCost` has been removed from the `AccountTransactionPayload` class.
--- a/examples/UpdateContract/Program.cs
+++ b/examples/UpdateContract/Program.cs
@@ -1,0 +1,95 @@
+using System.Globalization;
+using CommandLine;
+using Concordium.Sdk.Client;
+using Concordium.Sdk.Types;
+using Concordium.Sdk.Wallets;
+
+namespace Transactions.UpdateContract;
+
+// We disable these warnings since CommandLine needs to set properties in options
+// but we don't want to give default values.
+#pragma warning disable CS8618
+
+internal sealed class UpdateTransactionExampleOptions
+{
+    [Option(
+        'k',
+        "keys",
+        HelpText = "Path to a file with contents that is in the Concordium browser wallet key export format.",
+        Required = true
+    )]
+    public string WalletKeysFile { get; set; }
+    [Option(HelpText = "URL representing the endpoint where the gRPC V2 API is served.",
+        Default = "http://node.testnet.concordium.com:20000/")]
+    public string Endpoint { get; set; }
+    [Option('a', "amount", HelpText = "Amount of CCD to deposit.", Default = 0)]
+    public ulong Amount { get; set; }
+
+    [Option('c', "contract", HelpText = "The index of the smart contract.", Required = true)]
+    public ulong Contract { get; set; }
+
+    [Option('r', "receive-name", HelpText = "The receive_name of the contract to be called.", Required = true)]
+    public string ReceiveName { get; set; }
+
+    [Option('e', "max-energy", HelpText = "The maximum energy to spend on the module.", Required = true)]
+    public ulong MaxEnergy { get; set; }
+}
+
+/// <summary>
+/// Example demonstrating how to submit a transaction updating a smart contract.
+///
+/// The example assumes you have your account key information stored
+/// in the Concordium browser wallet key export format, and that a path
+/// pointing to it is supplied to it from the command line.
+///
+/// Cfr. <see cref="UpdateTransactionExampleOptions"/> for more info
+/// on how to run the program, or refer to the help message.
+/// </summary>
+internal class Program
+{
+    /// <summary>
+    /// Example send a contract update transaction.
+    /// </summary>
+    public static async Task Main(string[] args) =>
+        await Parser.Default
+            .ParseArguments<UpdateTransactionExampleOptions>(args)
+            .WithParsedAsync(Run);
+
+    private static async Task Run(UpdateTransactionExampleOptions options)
+    {
+        // Read the account keys from a file.
+        var walletData = File.ReadAllText(options.WalletKeysFile);
+        var account = WalletAccount.FromWalletKeyExportFormat(walletData);
+
+        // Construct the client.
+        var clientOptions = new ConcordiumClientOptions
+        {
+            Timeout = TimeSpan.FromSeconds(10)
+        };
+        using var client = new ConcordiumClient(new Uri(options.Endpoint), clientOptions);
+
+        // Create the update transaction.
+        var amount = CcdAmount.FromCcd(options.Amount);
+        var contract = ContractAddress.From(options.Contract, 0);
+        var receiveName = ReceiveName.Parse(options.ReceiveName);
+        var parameter = Parameter.Empty();
+        var maxEnergy = new EnergyAmount(options.MaxEnergy);
+
+        var updatePayload = new Concordium.Sdk.Transactions.UpdateContract(amount, contract, receiveName, parameter);
+
+        // Prepare the transaction for signing.
+        var sender = account.AccountAddress;
+        var sequenceNumber = client.GetNextAccountSequenceNumber(sender).Item1;
+        var expiry = Expiry.AtMinutesFromNow(30);
+        var preparedTransfer = updatePayload.Prepare(sender, sequenceNumber, expiry, maxEnergy);
+
+        // Sign the transaction using the account keys.
+        var signedTransfer = preparedTransfer.Sign(account);
+
+        // Submit the transaction.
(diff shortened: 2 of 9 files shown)


src/Types/Parameter.cs
```diff
@@ -1,10 +1,75 @@
...
+        }
+
+        var size = sizeRead + MinSerializedLength;
+        if (size > bytes.Length)
+        {
+            var msg = $"Invalid length of input in `Parameter.TryDeserial`. Expected array of size at least {size}, found {bytes.Length}";
+            output = (null, msg);
+            return false;
+        };
+        output = (new Parameter(bytes.Slice(sizeof(ushort), sizeRead).ToArray()), null);
```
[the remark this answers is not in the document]
limemloh: No, the bytes provided might include more than what we need.
We might not need that for this type currently, but this is needed by various other `TryDeserial` implementations.

src/Transactions/UpdateContract.cs
```diff
@@ -0,0 +1,124 @@
...
+        remaining_bytes = remaining_bytes[(int)CcdAmount.BytesLength..];
+
+        if (!ContractAddress.TryDeserial(remaining_bytes, out var address))
+        {
+            output = (null, address.Error);
+            return false;
+        };
+        remaining_bytes = remaining_bytes[(int)ContractAddress.BytesLength..];
+
+        if (!ReceiveName.TryDeserial(remaining_bytes, out var receiveName) || receiveName.receiveName == null)
```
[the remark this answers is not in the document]
limemloh: I need to know that it is not null already below this if statement, because I use it to create the span of remaining bytes.
It does stick out and looks annoying, though.

examples/UpdateContract/Program.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    [Option(
+        'k',
+        "keys",
+        HelpText = "Path to a file with contents that is in the Concordium browser wallet key export format.",
+        Required = true
+    )]
+    public string WalletKeysFile { get; set; }
+    [Option(HelpText = "URL representing the endpoint where the gRPC V2 API is served.",
+        Default = "http://node.testnet.concordium.com:20000/")]
```
abizjak, line 22: This URL will be deprecated soon, best to use https://grpc.testnet.concordium.com:20000 instead.

examples/UpdateContract/Program.cs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/// <summary>
+/// Example demonstrating how to submit a transaction updating a smart contract.
+///
+/// The example assumes you have your account key information stored
+/// in the Concordium browser wallet key export format, and that a path
+/// pointing to it is supplied to it from the command line.
+///
+/// Cfr. <see cref="UpdateTransactionExampleOptions"/> for more info
```
abizjak, line 44:
```suggestion
/// See <see cref="UpdateTransactionExampleOptions"/> for more info
```

examples/UpdateContract/Program.cs
```diff
@@ -0,0 +1,94 @@
...
+        var parameter = Parameter.Empty();
+        var maxEnergy = new EnergyAmount(options.MaxEnergy);
+
+        var updatePayload = new Concordium.Sdk.Transactions.UpdateContract(amount, contract, receiveName, parameter);
+
+        // Prepare the transaction for signing.
+        var sender = account.AccountAddress;
+        var sequenceNumber = client.GetNextAccountSequenceNumber(sender).Item1;
+        var expiry = Expiry.AtMinutesFromNow(30);
+        var preparedTransfer = updatePayload.Prepare(sender, sequenceNumber, expiry, maxEnergy);
```
abizjak, line 83: It's not exactly a transfer.

src/Transactions/UpdateContract.cs
```diff
@@ -0,0 +1,124 @@
...
+    ) => new(sender, sequenceNumber, expiry, new(_baseCost.Value + energy.Value), this);
+
+    /// <summary>
+    /// The transaction specific cost for submitting this type of
+    /// transaction to the chain.
+    ///
+    /// This should reflect the transaction-specific costs defined here:
+    /// https://github.com/Concordium/concordium-base/blob/78f557b8b8c94773a25e4f86a1a92bc323ea2e3d/haskell-src/Concordium/Cost.hs
+    /// </summary>
+    private static readonly EnergyAmount _baseCost = new(300);
```
abizjak, line 36: I don't understand where this is coming from. There is no such thing for updating contracts.

abizjak:
```rust
    /// Update a smart contract intance, giving it the given amount of energy
    /// for execution. The unique parameters are
    /// - `energy` -- the amount of energy that can be used for contract
    ///   execution. The base energy amount for transaction verification will be
    ///   added to this cost.
    pub fn update_contract(
        num_sigs: u32,
        sender: AccountAddress,
        nonce: Nonce,
        expiry: TransactionTime,
        payload: UpdateContractPayload,
        energy: Energy,
    ) -> PreAccountTransaction {
        let payload = Payload::Update { payload };
        make_transaction(
            sender,
            nonce,
            expiry,
            GivenEnergy::Add { num_sigs, energy },
            payload,
        )
    }
```

src/Types/ContractAddress.cs
```diff
@@ -20,10 +23,47 @@ public sealed record ContractAddress(ulong Index, ulong SubIndex) : IAddress
...
+    }
+
+    /// <summary>
+    /// Serialize the ContractAddress in big-endian format.
+    /// </summary>
+    public byte[] ToBytes()
+    {
+        using var memoryStream = new MemoryStream((int)BytesLength); // Safe to cast since we know BytesLength is within the range of int.
+        memoryStream.Write(Serialization.ToBytes(this.Index));
+        memoryStream.Write(Serialization.ToBytes(this.SubIndex));
```
abizjak, line 66: I'm probably just missing it, but I can't see where these Index/SubIndex types are defined so I can't see where it is defined that this will indeed be serialized as big endian.

src/Types/Parameter.cs
```diff
@@ -1,10 +1,75 @@
...
+
+        var sizeRead = BinaryPrimitives.ReadUInt16BigEndian(bytes);
+        if (sizeRead > MaxByteLength)
+        {
+            var msg = $"Invalid length of input in `Parameter.TryDeserial`. The parameter size can be at most {MaxByteLength} bytes, found {bytes.Length}";
+            output = (null, msg);
+            return false;
+        }
+
+        var size = sizeRead + MinSerializedLength;
```
abizjak, line 62: While I see that this is correct, it is conceptually odd, and I would probably prefer a 2 here.

src/Types/ReceiveName.cs
```diff
@@ -76,4 +100,73 @@ private static bool IsValid(string name, out ValidationError? error)
...
+    public static bool TryDeserial(ReadOnlySpan<byte> bytes, out (ReceiveName? receiveName, string? Error) output)
+    {
+        if (bytes.Length < MinSerializedLength)
+        {
+            var msg = $"Invalid length of input in `ReceiveName.TryDeserial`. Expected at least {MinSerializedLength}, found {bytes.Length}";
+            output = (null, msg);
+            return false;
+        };
+        var sizeRead = BinaryPrimitives.ReadUInt16BigEndian(bytes); // This should never throw, since we already checked the length.
+        var size = sizeRead + MinSerializedLength;
```
abizjak, line 118: Like for parameters, I think this is really an odd way of conceptualizing it. YOu hardcode UInt16 just in the line above, but then you have this constant MinSerializedLength for reasons I do not understand.

abizjak (approved):
Looks good apart from the oddness in the cost calculation.

CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
 ## Unreleased changes
   - Added
     - New GRPC-endpoint: `GetBlockItems`
-    - New transaction `DeployModule`
+    - New transactions: `DeployModule` and `UpdateContract`.
```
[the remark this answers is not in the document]
limemloh: Rasmus has a separate PR with that https://github.com/Concordium/concordium-net-sdk/pull/89

src/Types/ContractAddress.cs
limemloh: The fields are defined on line 15 of this file.
These are `ulong` and uses the serialization defined in [`Helpers/Serialization.cs`](https://github.com/Concordium/concordium-net-sdk/blob/main/src/Helpers/Serialization.cs).

src/Transactions/UpdateContract.cs
limemloh: This is from the base cost for updating instances, so the `updateContractInstanceBaseCost` in [Cost.hs](https://github.com/Concordium/concordium-base/blob/78f557b8b8c94773a25e4f86a1a92bc323ea2e3d/haskell-src/Concordium/Cost.hs).

abizjak: Yes, but I don't understand why this is added here. That cost is an internal implementation detail.

Whenever anybody supplies execution energy it will include this cost already. So we don't need to add it here.

src/Transactions/UpdateContract.cs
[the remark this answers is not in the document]
schwartz-concordium: Can’t you make it part of `ReceiveName.TryDeserial` to ensure it isn’t null and then use `!`. Seems wrong to me that you have the function but still needs to validate the output.

src/Transactions/UpdateContract.cs
limemloh: right... I'll remove this

schwartz-concordium (approved):

examples/UpdateContract/Program.cs
limemloh: I have updated all of the other examples as well

src/Transactions/UpdateContract.cs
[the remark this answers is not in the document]
limemloh: I have removed the check here

PR #93, opened 2024-01-26
FFI error handling

schwartz-concordium:
## Purpose

Returns result with possible error codes from FFI calls.

## Changes

- Added better error handling from FFI by returning error code indicating the type of error
- Make exception with error code public such that it can be caught by clients

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-bindings/src/lib.rs
+++ b/rust-bindings/src/lib.rs
@@ -49,7 +49,7 @@ pub unsafe extern "C" fn schema_display(
     schema_size: i32,
     schema_version: FFIByteOption,
     callback: ResultCallback,
-) -> u32 {
+) -> u8 {
     let schema = std::slice::from_raw_parts(schema_ptr, schema_size as usize);
     assign_result(callback, || {
         schema_display_aux(schema, schema_version.into_option())
@@ -88,7 +88,7 @@ pub unsafe extern "C" fn get_receive_contract_parameter(
     value_ptr: *const u8,
     value_size: i32,
     callback: ResultCallback,
-) -> u32 {
+) -> u8 {
     assign_result(callback, || {
         let schema = std::slice::from_raw_parts(schema_ptr, schema_size as usize);
         let contract_name_str = get_str_from_pointer(contract_name)?;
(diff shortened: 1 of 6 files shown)


schwartz-concordium:
Hi @limemloh 


I have made this PR with better error mapping of FFI errors. I’m not sure the interface given to developers are 100 %, so can you give it a review and then we can have a discussion if we can make some improvements.

rust-bindings/src/lib.rs
```diff
@@ -182,45 +184,90 @@ pub fn get_receive_contract_parameter_aux(
...
+    #[error("error when using serde")]
+    SerdeJsonError,
+    #[error("encountered string which wasn't utf8 encoded")]
+    Utf8Error,
+    #[error(transparent)]
+    VersionedSchemaError(#[from] VersionedSchemaError),
+}
+
+impl FFIError {
+    fn to_int(&self) -> u16 {
```
limemloh, line 212: **Suggestion:** It would be good with a comment here to why it is not starting from `0`

src/Exceptions/InteropBindingException.cs
```diff
@@ -1,20 +1,23 @@
 using System.Text;
+using Concordium.Sdk.Interop;
 
 namespace Concordium.Sdk.Exceptions;
 
 /// <summary>
 /// Thrown when a interop call failed with possible error as message.
 /// </summary>
-internal sealed class InteropBindingException : Exception
+public sealed class InteropBindingException : Exception
```
limemloh, line 9: **Discussion:** If we are making this public, we should probably rename it, since the current name is more related to implementation details.
Maybe `SchemaJsonException`?

src/Interop/Result.cs
```diff
@@ -0,0 +1,91 @@
+namespace Concordium.Sdk.Interop;
+
+/// <summary>
+/// Result type which on errors hold error type information.
+/// </summary>
+public enum Result
```
limemloh, line 6: **Suggestion:** The name should probably be more specific.
```suggestion
public enum SchemaJsonResult
```

limemloh (approved):

PR #95, opened 2024-02-07
Support JSON parameters

### LINKED ISSUE #94: Support JSON (with schema) for constructing parameters
**Task description**

Support using a smart contract schema to construct smart contract parameters, for both `init` and `update` transactions.

**Sub-tasks**
- [x] Provide a method on the `Parameter` class to construct it from a schema and JSON.
- [x] Provide an example using the above method to mint some CIS-2 tokens, confirming the minting by checking the resulting events from the transaction.

limemloh:
## Purpose

Support smart contract parameters in JSON format using the smart contract module schema.
Closes #94 

## Changes

- Add `WaitUntilFinalized` method on `ConcordiumClient` for waiting for transactions to finalize.
- Add `Parameter.UpdateJson` and `Parameter.InitJson` for constructing parameters for update and init transactions using JSON and smart contract module schemas.
- Minor changes to `FFIError` in `lib.rs` removing some manual `From` implementations.

## Checklist

      hard-to-understand areas.


examples/UpdateContractMint/Program.cs
```diff
@@ -0,0 +1,260 @@
...
+/// - Submitting a transaction updating a smart contract, where the parameter is constructed
+///   using a smart contract module schema.
+/// - Waiting for the transaction to finalize.
+/// - Checking the outcome and reading the logged events of the contract, parsing the events using
+///   a smart contract module schema.
+///
+/// The example assumes:
+/// - You have your account key information stored in the Concordium browser wallet key export
+///   format, and that a path pointing to it is supplied to it from the command line.
+/// - The 'cis2_multi' smart contract example is deployed on chain (already on testnet) with a
```
EmilLa1, line 52: I believe that another prerequisite is that the contract MUST be initialized - otherwise where should the contract address stem from.

examples/UpdateContractMint/Program.cs
```diff
@@ -0,0 +1,260 @@
...
+        Console.WriteLine($"Transaction got accepted!");
+
+        // Exact the logs of updated smart contract instances from the outcome.
+        if (!finalized.Summary.TryGetContractUpdateLogs(out var updates))
+        {
+            throw new InvalidOperationException(
+                "Transaction summary failed to parse as a contract update transaction."
+            );
+        }
+        // Print out the events from each updated contract.
```
EmilLa1, line 169:
```suggestion
        // Print out the events from each updated contract in the block.
```

examples/UpdateContractMint/Program.cs
```diff
@@ -0,0 +1,260 @@
...
+            if (updatedContract == contract)
+            {
+                foreach (var evt in update.Item2)
+                {
+                    // Use the smart contract module schema to deserialize the events.
+                    var json = evt.GetDeserializeEvent(moduleSchema, receiveName.GetContractName());
+                    Console.WriteLine($"- {json}");
+                }
+            }
+            else
```
EmilLa1, line 184: Are other contracts relevant?

examples/UpdateContractMint/Program.cs
```diff
@@ -0,0 +1,260 @@
...
+    {
+        /// <summary>
+        /// The owner for the newly minted tokens.
+        ///
+        /// This property only represents the owner being an account address, but the smart contract
+        /// also allow for a contract address here, which we choose not include in the type for
+        /// simplicity.
+        /// </summary>
+        [JsonPropertyName("owner")]
+        public JsonAccountAddr Owner { get; set; }
```
EmilLa1, line 210: I get the simplicity part. 
So this is more of a question. Don't we have "proper" types reflecting these? 

There's quite a lot of invariants on these types, e.g. the address is 32 bytes, the format of the token id etc.

examples/UpdateContractMint/Program.cs
```diff
@@ -0,0 +1,260 @@
...
+        [JsonPropertyName("hash")]
+        public JsonNoHash ContentHash { get; set; }
+    }
+
+    /// <summary>
+    /// Indicator of no hash provided for the token metadata.
+    /// </summary>
+    private class JsonNoHash
+    {
+        public string[] None { get; set; }
```
EmilLa1, line 258: Why is this a string array?

examples/UpdateContractMint/Program.cs
```diff
@@ -0,0 +1,260 @@
...
+        ///
+        /// This property only supports not providing a content hash of the metadata, but the smart
+        /// contract supports this as well, we choose not include this in the type for simplicity.
+        /// </summary>
+        [JsonPropertyName("hash")]
+        public JsonNoHash ContentHash { get; set; }
+    }
+
+    /// <summary>
+    /// Indicator of no hash provided for the token metadata.
```
EmilLa1, line 254:
```suggestion
    /// Indicator of no checksum (SHA256 hash) provided for the token metadata.
```

EmilLa1 (approved):
Looks great!

Minor comments - probably most because of my ignorance of C#, so take it for what it is :-).

src/Client/ConcordiumClient.cs
```diff
@@ -205,6 +205,52 @@ public ITransactionStatus GetBlockItemStatus(TransactionHash transactionHash, Ca
...
+                    output = finalized.State;
+                    return true;
+                default:
+                    output = null;
+                    return false;
+            };
+        }
+
+        // Check the transaction status straight away and return if finalized already.
+        var status = await this.GetBlockItemStatusAsync(transactionHash, token);
```
abizjak, line 234: If the transaction is missing, does this raise an exception or return "not found"?

src/Types/Parameter.cs
```diff
@@ -33,14 +33,41 @@ public sealed record Parameter(byte[] Param) : IEquatable<Parameter>
...
 
+    /// <summary>
+    /// Create a parameter from JSON representation using the smart contract module schema for a smart contract update transaction.
+    /// </summary>
+    /// <param name="moduleSchema">The smart contract module schema.</param>
+    /// <param name="contractName">The name of the contract.</param>
+    /// <param name="functionName">The name of entrypoint of the smart contract.</param>
+    /// <param name="jsonParameter">The UTF8 encoding of the JSON representation of the smart contract parameter.</param>
+    public static Parameter UpdateJson(
+        VersionedModuleSchema moduleSchema,
```
abizjak, line 52: Is there a reason why parameter schemas are not supported?

We advertise those since they allow you to interact with a "generic CIS2 contract".

examples/UpdateContractMint/Program.cs
limemloh: No not really. Initially I just assumed our contract is the only one, but then I had to provide a complex comment explaining this and what to do when it is not the case, so I choose instead to make the example more general.

examples/UpdateContractMint/Program.cs
limemloh: We discussed this offline, but for anyone else reading this:
There are proper types defined for the different types, but these do not provide a way to generate the correct JSON. In the future, we might provide this, but not part of this PR.

examples/UpdateContractMint/Program.cs
limemloh: The JSON needs to be `{ "None": [] }`, so any array will do.
I have extended on this in the comment

src/Client/ConcordiumClient.cs
limemloh: It will be an RpcException with status of not found

abizjak: Ok, there is a bit of a race condition in common usage "send tx + wait for finalization" so it'd be useful to document this behavior that if the transaction is not yet there when this is called it'll fail.

src/Types/Parameter.cs
limemloh: I have added a `FromJson` as well, which takes a schema type and not the whole module schema.

abizjak (approved):

PR #101, opened 2024-09-11
Fix build warnings

### LINKED ISSUE #65: Remove documentation build warnings
Since [updating the configuration file](https://github.com/Concordium/concordium-net-sdk/pull/64) to include documentation, several warnings for missing documentation and formatting are now shown during compilation.

We need to fix all the warnings and ideally have the CI build [treat warnings as errors](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#treatwarningsaserrors) for future PRs.

limemloh:
## Purpose

Closes #65

Building the project produces a lot of warnings, most are about missing documentation and end up hiding the more important ones.
This PR fixes every warning, but does not include full support for Concordium node v7 API, this was [added here instead](https://github.com/Concordium/concordium-net-sdk/pull/102)

### DIFF
--- a/src/Types/BakerEvent.cs
+++ b/src/Types/BakerEvent.cs
@@ -51,6 +51,7 @@ internal static class BakerEventFactory
                 ),
             BakerEvent.EventOneofCase.None =>
                 throw new MissingEnumException<BakerEvent.EventOneofCase>(bakerEvent.EventCase),
+            BakerEvent.EventOneofCase.DelegationRemoved => throw new NotImplementedException(),
             _ => throw new MissingEnumException<BakerEvent.EventOneofCase>(bakerEvent.EventCase)
         };
 }
--- a/src/Types/BlockHashInput.cs
+++ b/src/Types/BlockHashInput.cs
@@ -7,6 +7,9 @@ namespace Concordium.Sdk.Types;
 /// </summary>
 public interface IBlockHashInput
 {
+    /// <summary>
+    /// Convert this into type expected by the gRPC.
+    /// </summary>
     BlockHashInput Into();
 }
 
@@ -17,6 +20,9 @@ public sealed record Best : IBlockHashInput
 {
     private static readonly Empty _empty = new();
 
+    /// <summary>
+    /// Convert this into type expected by the gRPC.
+    /// </summary>
     public BlockHashInput Into() => new()
     {
         Best = _empty,
@@ -31,6 +37,9 @@ public sealed record LastFinal : IBlockHashInput
 {
     private static readonly Empty _empty = new();
 
+    /// <summary>
+    /// Convert this into type expected by the gRPC.
+    /// </summary>
     public BlockHashInput Into() => new()
     {
         LastFinal = _empty
@@ -42,6 +51,9 @@ public sealed record LastFinal : IBlockHashInput
 /// </summary>
 public sealed record Given(BlockHash BlockHash) : IBlockHashInput
 {
+    /// <summary>
+    /// Convert this into type expected by the gRPC.
+    /// </summary>
     public BlockHashInput Into() => new() { Given = this.BlockHash.ToProto() };
 }
 
--- a/src/Types/BlockHashInput.cs
+++ b/src/Types/BlockHashInput.cs
@@ -8,7 +8,7 @@ namespace Concordium.Sdk.Types;
 public interface IBlockHashInput
 {
     /// <summary>
-    /// Convert this into type expected by the gRPC.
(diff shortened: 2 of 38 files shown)


src/Types/BakerEvent.cs
```diff
@@ -51,6 +51,7 @@ internal static IBakerEvent From(BakerEvent bakerEvent) =>
                 ),
             BakerEvent.EventOneofCase.None =>
                 throw new MissingEnumException<BakerEvent.EventOneofCase>(bakerEvent.EventCase),
+            BakerEvent.EventOneofCase.DelegationRemoved => throw new NotImplementedException(),
```
Victor-N-Suadicani, line 54: Why? A comment to explain why this isn't implemented would be nice.

src/Types/BlockHashInput.cs
```diff
@@ -7,6 +7,9 @@ namespace Concordium.Sdk.Types;
 /// </summary>
 public interface IBlockHashInput
 {
+    /// <summary>
+    /// Convert this into type expected by the gRPC.
```
Victor-N-Suadicani, line 11: Perhaps:
```suggestion
    /// Convert this into the type expected by the gRPC API.
```

src/Types/BlockHeight.cs
```diff
@@ -40,6 +46,9 @@ public BlocksAtHeightRequest Into() =>
 /// </param>
 public sealed record RelativeHeight(ulong Height, uint GenesisIndex, bool Restrict) : IBlockHeight
 {
+    /// <summary>
+    /// Convert an RelativeHeight into a request for blocks at height.
```
Victor-N-Suadicani, line 50:
```suggestion
    /// Convert a RelativeHeight into a request for blocks at height.
```

Also request for blocks at what height? I guess I'm missing some context but the sentence just sounds weird. Same in the other places.

src/Types/CcdAmount.cs
```diff
@@ -136,10 +136,27 @@ public static bool TryDeserial(ReadOnlySpan<byte> bytes, out (CcdAmount? Amount,
...
+    public static CcdAmount Max(CcdAmount first, CcdAmount second)
+    {
+        if (first < second)
+        {
+            return second;
+        }
+        else
+        {
+            return first;
+        }
```
Victor-N-Suadicani, line 151: Does this not work? (maybe you need an explicit `return` in front, not sure)
```suggestion
        first < second ? second : first
```

src/Types/DelegationEvent.cs
```diff
@@ -27,6 +27,7 @@ internal static IDelegationEvent From(DelegationEvent delegationEvent) =>
                 DelegationAdded.From(delegationEvent.DelegationAdded),
             DelegationEvent.EventOneofCase.DelegationRemoved =>
                 DelegationRemoved.From(delegationEvent.DelegationRemoved),
+            DelegationEvent.EventOneofCase.BakerRemoved => throw new NotImplementedException(),
```
Victor-N-Suadicani, line 30: Same here, why is this not implemented?

src/Types/UpdatePayload.cs
```diff
@@ -16,28 +16,51 @@ public interface IUpdatePayload
 /// </summary>
 public enum UpdateType
 {
+    /// <summary> Update of protocol version. </summary>
```
Victor-N-Suadicani, line 19: I love doc comments more than the average person probably but a lot of these doc comments really add nothing. Is there nothing more to add? Or could we not disable whatever requirement is forcing these i.e. akin to Rust's `#[allow(missing_docs)]`?

Victor-N-Suadicani (approved):
Look alright, just a few comments but nothing blocking.

src/Types/BakerEvent.cs
limemloh: I tried to split up changes related to fixing build warnings and changes due to P7.
It is implemented in https://github.com/Concordium/concordium-net-sdk/pull/102

src/Types/BlockHeight.cs
limemloh: I change the doc comment to make it more clear

src/Types/UpdatePayload.cs
limemloh: I don't think there is a lot more to add to this.
To me, this still seems better than disabling the warning.

src/Types/DelegationEvent.cs
limemloh: It is implemented in https://github.com/Concordium/concordium-net-sdk/pull/102

